PR #6, opened 2018-03-23
update switch sdl2

Cpasjuste:


WinterMute:
need to update the sha256sum - https://github.com/devkitPro/pacman-packages/blob/master/switch/SDL2/PKGBUILD#L15

PR #10, opened 2018-03-25
add switch libmodplug

Cpasjuste:


Cpasjuste:
Not ready yet : /home/cpasjuste/dev/multi/pacman-packages/switch/libmodplug/src/libmodplug-0.8.8.5/src/load_pat.cpp:1456: undefined reference to `sleep'
Fixing soon !

Cpasjuste:
Should i just patch "sleep(x)" to "usleep(x*1000*1000)" just to fix linking ? We said the sleep case should mainly never be used.

PR #11, opened 2018-03-26
add switch libtremor

Cpasjuste:


switch/libtremor/PKGBUILD
```diff
@@ -0,0 +1,38 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+# Contributor: cpasjuste <cpasjuste@gmail.com>
+
+pkgname=switch-libtremor
+pkgver=9999
```
carstene1ns: You do not really want to do this.

WinterMute:
We have libvorbis, I don't believe we also need or want this.

The source you've used is also improperly packaged, see https://github.com/devkitPro/pacman-packages/tree/master/3ds/libvorbisidec

Cpasjuste:
I do need it at least. I'm no going to refactor the ports using it :/

carstene1ns:
@Cpasjuste At least tremor is nearly api compatible with ogg+vorbis(file).
Apart from build system stuff this works by just changing the headers and hardcoding a bit:
https://github.com/EasyRPG/Player/blob/master/src/decoder_oggvorbis.cpp#L119-L123

Ghabry:
at least I observed during testing that libtremor is much faster in decoding than libvorbis.

WinterMute:
Using correct packaging here https://github.com/devkitPro/pacman-packages/commit/ad6490250a1143d952693e22a950683593793d8d

PR #12, opened 2018-03-26
Add mpg123 1.25.10 as portlib for 3ds, ppc and switch

Ghabry:
Best library for mp3 playback.

The patch disables the unused player cli interface (needs command line stuff) and network functionality.

No idea what the network code does, I guess MPEG file streaming from a website?
Though I created this port years ago when there was no socket code in newlib so can't say if it works by now. (Patches welcome if anybody really needs this)

Dependency for SDL2 mixer if anybody wants to port this :+1:

### DIFF
--- a/3ds/mpg123/PKGBUILD
+++ b/3ds/mpg123/PKGBUILD
@@ -1,0 +1,44 @@
+# Maintainer: Gabriel Kind <gabriel-dkp mastergk de>
+
+pkgname=3ds-mpg123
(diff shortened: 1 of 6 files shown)


WinterMute:
Used a single patch file for all 3 and added pkg-config to the build deps for these https://github.com/devkitPro/pacman-packages/commit/62935334376aec9c8cd28c3a45a1d6f3f8060ba9

PR #14, opened 2018-03-30
add switch libmad

Cpasjuste:


switch/libmad/PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+source=("ftp://ftp.mars.org/pub/mpeg/libmad-${pkgver}.tar.gz")
+sha256sums=('bbfac3ed6bfbc2823d3775ebb931087371e142bb0e9bb1bee51a76a6e0078690')
+
+build() {
+  cd libmad-$pkgver
+
+  source /opt/devkitpro/devkita64.sh
+  source /opt/devkitpro/switchvars.sh
+
+  # add aarch64 support
```
WinterMute: This is bogus. Replace config.sub and config.guess with more recent versions from your  local /usr/share/automake-* & generate a patch

switch/libmad/PKGBUILD
```diff
@@ -0,0 +1,40 @@
...
+  sed -i '230i | aarch64 | aarch64_be \\' config.sub
+  sed -i '296i | aarch64-* | aarch64_be-* \\' config.sub
+
+  ./configure --prefix="$DEVKITPRO/portlibs/switch" --host=aarch64-none-elf \
+    --disable-shared --enable-static --enable-speed
+
+  # remove invalid flags
+  sed  -i 's/-fforce-mem//g' Makefile
```
WinterMute: This is just bodgery. Make sure the configury is right then look for options to disable this kind of thing in the configure script

WinterMute:
Had to do some more involved patching of configury. Did it this way - https://github.com/devkitPro/pacman-packages/commit/714710de2980cef3751e3adea37a6d0c90fb37a1

PR #17, opened 2018-04-03
Fix another typo in 3ds SDL port

Pirater12:


Pirater12:
Looks lime brnch was not up to date

PR #23, opened 2018-04-19
3ds/mbedtls: Update to 2.8.0

Exagone313:
Release notes: https://tls.mbed.org/tech-updates/releases/mbedtls-2.8.0-2.7.2-and-2.1.11-released

### DIFF
--- a/3ds/mbedtls/PKGBUILD
+++ b/3ds/mbedtls/PKGBUILD
@@ -1,7 +1,8 @@
-
 # Maintainer: WinterMute <davem@devkitpro.org>
+# Contributor: Elouan Martinet <exa@elou.world>
+
 pkgname=3ds-mbedtls
-pkgver=2.7.0
+pkgver=2.8.0
 pkgrel=1
 pkgdesc='Deflate compression method library (for Nintendo 3DS homebrew development)'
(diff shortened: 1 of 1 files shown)


3ds/mbedtls/PKGBUILD
```diff
@@ -10,12 +11,11 @@ license=('apache')
 options=(!strip libtool staticlibs)
 depends=(3ds-zlib)
 makedepends=('3ds-pkg-config' 'devkitpro-pkgbuild-helpers')
-source=( 'https://tls.mbed.org/download/mbedtls-2.7.0-apache.tgz'
+source=( "https://tls.mbed.org/download/mbedtls-${pkgver}-apache.tgz"
          'libmbedtls-2.7.0.patch' )
```
WinterMute: If this patch applies clean (i.e. doesn't fuzz to find the right lines) then rename it & use ${pkgver} for this too.

If it needs to fuzz then generate a new patch for 2.8.0

WinterMute:
I reverted this again. The build failed, don't really have time to look at why right now.

Exagone313:
Hmm, it works locally.

PR #31, opened 2018-06-11
Added libuuid and wolfSSL Switch portlibs

natinusala:
wolfSSL is pulled from the official GitHub repository with a patch applied to use `csrng` as a seed generator.

I wasn't able to build just libuuid from linux-utils (gettext was not friendly), so I pulled it from a SourceForge repository and adapted ; the final code is hosted there : https://github.com/natinusala/switch-libuuid (it doesn't use uuidd)

libuuid **is not thread-safe** as I removed the internal clock file lock due to missing `flock` function.

### DIFF
--- a/switch/libuuid/.gitignore
+++ b/switch/libuuid/.gitignore
@@ -1,0 +1,1 @@
+switch-libuuid
--- a/switch/libuuid/PKGBUILD
+++ b/switch/libuuid/PKGBUILD
@@ -6,6 +6,7 @@ url='https://sourceforge.net/projects/libuuid'
 license=('GPL-2.0')
 options=(!strip libtool staticlibs)
 source=("https://github.com/natinusala/switch-libuuid/archive/$pkgver.zip")
+groups=('switch-portlibs')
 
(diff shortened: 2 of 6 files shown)


WinterMute:
Did you  get anywhere with mbedtls?

I'd prefer these PKGBUILDs pull original source & patch where possible. It makes things much easier to maintain & update in the future.

natinusala:
The EPIPE / kernel error issue is still present. However, kbhomes managed to make OpenSSL work (apparently without hardware seeding) : https://github.com/kbhomes/moonlight-switch/tree/master/dependencies

Anyway this PR should be discarded as WolfSSL is no good

WinterMute:
Can you share your mbedtls patches somewhere so other people can have a look? Ideally we'd like to get that EPIPE issue fixed rather than just not porting things because it's still malingering in there.

I looked at the moonlight port and so far I've mainly been sucking air through my teeth.

natinusala:
Everything is there : https://github.com/natinusala/pacman-packages/tree/mbedtls/switch/mbedtls

natinusala:
This is the epipe issue. In the meantime you can use OpenSSL, look at moonlight-switch.

PR #34, opened 2018-06-25
Add libfribidi and libass for Switch

carstene1ns:
Some notes:
Currently `libass` with `--disable-require-system-font-provider` is not really usable OOTB, because font finding and selection will not work automatically. However, fontconfig might be a bit much to have. Thought about hardcoding the shared_font or something, but this can also be done from the app using it.

Btw. I do not know if you you intend to build packages from OSX, since `sed` usage in fribidi-prepare() is not portable. Might be worth a note in the README that real `.patch`es are preferred.

### DIFF
--- a/switch/libass/PKGBUILD
+++ b/switch/libass/PKGBUILD
@@ -1,0 +1,42 @@
+# Maintainer: Dave Murphy <davem@devkitpro.org>
+# Contributor: carstene1ns <dev f4ke de>
+
+pkgname=switch-libass
+pkgver=0.14.0
+pkgrel=1
+pkgdesc='A portable subtitle renderer (Nintendo Switch port)'
+arch=('any')
+url="https://github.com/libass/libass"
+license=('custom: ISC')
+options=(!strip staticlibs)
+depends=('switch-freetype' 'switch-libfribidi')
+makedepends=('switch-pkg-config' 'devkitpro-pkgbuild-helpers')
+source=("https://github.com/libass/libass/releases/download/$pkgver/libass-$pkgver.tar.xz")
+sha256sums=('881f2382af48aead75b7a0e02e65d88c5ebd369fe46bc77d9270a94aa8fd38a2')
+groups=('switch-portlibs')
+
+build() {
+  cd libass-$pkgver
+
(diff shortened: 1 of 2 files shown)


WinterMute:
Patching Makefile.am also requires automake which won't work on vanilla OSX any more - apple have removed autoconf/automake :/

Do the binaries not build at all or can they just be removed from the package at the end?

Ideally I'd like to see changes in upstream packages to allow disabling binaries that won't ever be used in some environments but perhaps that's hoping for too much.

WinterMute:
Ugh, this isn't building & I have no idea how to re-open the PR.

patching Makefile.am on the fly is doing weird stuff on my build machines.

```
 cd . && /bin/bash /home/davem/projects/devkitpro/pacman-packages/switch/libfribidi/src/fribidi-1.0.4/missing automake-1.15 --gnits Makefile
configure.ac:41: error: version mismatch.  This is Automake 1.15,
configure.ac:41: but the definition used by this AM_INIT_AUTOMAKE
configure.ac:41: comes from Automake 1.15.1.  You should recreate
configure.ac:41: aclocal.m4 with aclocal and run automake again.
WARNING: 'automake-1.15' is probably too old.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
Makefile:413: recipe for target 'Makefile.in' failed
make: *** [Makefile.in] Error 63
```

configure,ac is claiming 1.11.1 so I can only assume that's a lie & the release tarballs have been generated with 1.15.1 :(

carstene1ns:
meh :/ Is this fixed by 27512aa8 ?
If not, i can write a different patch (actually compiling the binaries) and submit that.

carstene1ns:
Yeah, they only exclude it when using glibc it seems, wanted to patch that out.

PR #35, opened 2018-06-26
Added switch-libxml2.

DarkMatterCore:


WinterMute:
The patch for this is wrong. Changing the include path will break anything that uses it.

libxml2 also seems rather serious overkill. Is there some reason to have it as well as [switch-expat](https://github.com/devkitPro/pacman-packages/tree/master/switch/libexpat)?

DarkMatterCore:
libxml2 is a lot easier to work with of you're handling big datasets (like the one from nswdb.com). libexpat is fine, but requires more work and refining for this kind of tasks. It's better to have two different flavors and let people decide what to use.

Leaving the include path the way it is actually triggers some problems while linking against the compiled library. No problems at all while compiling the library itself.

I'm already working on a project that uses libxml2: https://github.com/DarkMatterCore/gcdumptool. Working fine so far.

WinterMute:
Changing the include path can't possibly cause or fix linking issues ...

If the compiler can't find the includes when you're compiling it's because you're using the wrong include path. Note the output of 
```
/opt/devkitpro/portlibs/switch/bin/aarch64-none-elf-pkg-config libxml-2.0 --cflags
```

Can you give me an example of the "more work and refining" that using libexpat means?

DarkMatterCore:
If the include path is left unmodified, header files are copied to:

/opt/devkitpro/portlibs/switch/include/libxml2/libxml/

Trying to include a header file from that directory using, for example:

#include <libxml2/libxml/globals.h>

Results in errors at the linking stage for the application because of the way the header files internally point to each other. Changing the include directory is faster (instead of modifying the headers themselves) and doesn't cause problems.

libexpat is stream-oriented and puts more emphasys on XML tags rather than on their contents. Parsing the contents requires more work. I was able to do a simple parser with libxml2 using XPath queries and very few lines of code.

WinterMute:
No errors occur at the linking stage ...

Like I said - note the output of 
```
/opt/devkitpro/portlibs/switch/bin/aarch64-none-elf-pkg-config libxml-2.0 --cflags
```

Don't `"fix" libraries so your broken code works. Fix your broken code so it uses the library properly.

Can you tell me more about "parsing  contents" - that sounds kind of wrong tbh. It seems rather bizarre to be xml tagging data and then needing to parse that data further.

DarkMatterCore:
I'm on Windows, any calls to /opt/devkitpro/portlibs/switch/bin/aarch64-none-elf-pkg-config using the msys2 bundled with devkitPro (with or without args) fail and return this error:

línea 11: exec: pkg-config: no se encontró

Which roughly translates to "not found". Any help?

Compiling gcdumptool without changing the include directory from libxml2 and using these #include in my source code:

#include <libxml2/libxml/globals.h>
#include <libxml2/libxml/xpath.h>

Makes the compiler complain. Here's the log:


aarch64-none-elf-gcc -MMD -MP -MF /c/GitHub/gcdumptool/build/util.d -g -Wall -O2 -ffunction-sections -march=armv8-a -mtune=cortex-a57 -mtp=soft -fPIE  -I/c/GitHub/gcdumptool/include -I/opt/devkitpro/portlibs/switch/include -I/opt/devkitpro/libnx/include -I/c/GitHub/gcdumptool/build -D__SWITCH__ -c /c/GitHub/gcdumptool/source/util.c -o util.o
In file included from C:/GitHub/gcdumptool/source/util.c:11:
C:/devkitPro/portlibs/switch/include/libxml2/libxml/globals.h:17:10: fatal error: libxml/xmlversion.h: No such file or directory
 #include <libxml/xmlversion.h>
          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[1]: *** [/opt/devkitpro/devkitA64/base_rules:19: util.o] Error 1
make: *** [Makefile:157: build] Error 2


So, how can I fix that without changing the include dir or messing with the header files?

About parsing contents, my application needs to parse data off the XML "database" from nswdb.com to perform XCI dump comparison. You can find it at [http://nswdb.com/xml.php](url).

WinterMute:
`pacman -S pkg-config` then try again, sorry

DarkMatterCore:
Thanks, this is the output I get:

-I/opt/devkitpro/portlibs/switch/include/libxml2

You're right, the patch is wrong. I'm the one who should be saying sorry.

Do you have any suggestions? I was thinking about patching the libxml-2.0.pc.in file as well, but I don't know if you'd be OK with that. I really think this could be an useful package for some.

WinterMute:
I was hoping it might be reasonably obvious from the output of that command.
add
```
`aarch64-none-elf-pkg-config libxml-2.0 --cflags`
```
to your CFLAGS in the Makefile here https://github.com/DarkMatterCore/gcdumptool/blob/master/Makefile#L55

Are there build failures without --without-html, --without-ftp, and --without-http?

DarkMatterCore:
Will do, thanks. I'll also remove the patch from my fork.

And yes, a lot of failures actually. A lot of the code makes references to network functions that are not available in libnx nor in the standard libraries bundled with devkitA64 (like getaddrinfo()). Fixing this would require a lot of changes to the library's code, and given that the network stack from libnx is still being developed, I thought it would be a good idea to just disable those features.

WinterMute:
OK, cool. That worked.

The functions *are* available in libnx, you just need to link with it to get configure to understand that they're there.

I also removed some extra binaries that won't work on the host and realised that there's an xml2-config in there too so adding 
```
`xml2-config --cflags`
```
to CFLAGS will do the same thing as the pkg-config line we used before.

Currently this needs some changes in libnx that we haven't pushed out to a stable release yet so I won't merge this just yet.

DarkMatterCore:
Hi there. I just uploaded a new commit to also add the switch-libjson-c package. Didn't give me any trouble at all to compile and get working. I needed it to parse JSON responses using libcurl, may as well be useful for other people.

WinterMute:
Can you revert this to https://github.com/devkitPro/pacman-packages/pull/35/commits/c10032368a37862df593575d0d8e41acc09b1fc6

Merging master into a topic branch makes a mess that can be difficult to unravel.

I'd like one package per PR too, I'm not going to be accepting any more PRs with more than one package.

DarkMatterCore:
Sure, I can. I'll do it in a few minutes. Thanks for the heads up, and sorry for the trouble.

WinterMute:
Don't apologise, it's really not that much trouble.

DarkMatterCore:
It's done. Sorry for taking so long, I was having a lot of problems with my ISP. It's a common thing.

PR #36, opened 2018-06-27
Add libopus

kbhomes:
libopus builds cleanly and works as expected without changes.


Pirater12:
Could you add it for 3ds too?

PR #43, opened 2018-08-09
Add opusfile

Oreo639:
Requires libopus. Also was apart of dkp-portlibs but was never carried over.

### DIFF
--- a/3ds/opusfile/PKGBUILD
+++ b/3ds/opusfile/PKGBUILD
@@ -1,0 +1,32 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+
+pkgname=3ds-opusfile
+pkgver=0.10
+pkgrel=1
+pkgdesc='Library for opening, seeking, and decoding .opus files'
+arch=('any')
+url='https://opus-codec.org/'
+license=(BSD)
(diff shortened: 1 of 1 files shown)


3ds/opusfile/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+source=("https://downloads.xiph.org/releases/opus/opusfile-$pkgver.tar.gz")
+sha256sums=('48e03526ba87ef9cf5f1c47b5ebe3aa195bd89b912a57060c36184a6cd19412f')
+makedepends=('3ds-pkg-config' 'devkitpro-pkgbuild-helpers' '3ds-libogg' '3ds-libopus')
+
+build() {
+  cd opusfile-$pkgver
+
+  source /opt/devkitpro/3dsvars.sh
+
+  autoreconf -fi
```
WinterMute: autoreconf isn't needed here, nothing has been patched.

Ideally we want to avoid regenerating configury - it can cause maintenance issues.

Oreo639: Ops, I forgot to remove that when I copped libopus's PKGBUILD. Thank you for pointing it out.

PR #45, opened 2018-09-11
Add enet for the switch

Subv:
The enet source is cloned from https://github.com/lsalzman/enet , specifically this commit https://github.com/lsalzman/enet/commit/e2ef83927d1626d6ee479742343a3011333fdfdd since there last released version (1.3.13) is more than 3 years old and contains several include issues that prevent it from building for the switch.

### DIFF
--- a/switch/enet/PKGBUILD
+++ b/switch/enet/PKGBUILD
@@ -1,0 +1,39 @@
+pkgname=switch-enet
+pkgver=r252.e2ef839
+pkgrel=1
+pkgdesc='ENet reliable UDP networking library (adapted for Nintendo Switch homebrew development)'
+arch=('any')
+url='https://github.com/lsalzman/enet'
+license=('MIT')
+options=(!strip libtool staticlibs)
+makedepends=('git' 'switch-pkg-config' 'devkitpro-pkgbuild-helpers')
+source=(${pkgname}::"git+https://github.com/lsalzman/enet.git#commit=e2ef83927d1626d6ee479742343a3011333fdfdd")
+md5sums=('SKIP')
+
+pkgver() {
+    cd "$srcdir/${pkgname}"
+    printf 'r%s.%s' "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)"
+}
+
+build() {
+  cd ${pkgname}
+
+  source /opt/devkitpro/devkita64.sh
+  source /opt/devkitpro/switchvars.sh
+
+  autoreconf -vfi
+
+  ./configure --prefix=$PORTLIBS_PREFIX --host=aarch64-none-elf \
+    --disable-shared --enable-static
+
+  make
+}
+
+package() {
+  cd ${pkgname}
+
+  source /opt/devkitpro/devkita64.sh
+  source /opt/devkitpro/switchvars.sh
+
+  make DESTDIR="$pkgdir" install
+}
--- a/switch/enet/PKGBUILD
+++ b/switch/enet/PKGBUILD
@@ -1,24 +1,18 @@
 pkgname=switch-enet
-pkgver=r252.e2ef839
+pkgver=1.3.15
 pkgrel=1
 pkgdesc='ENet reliable UDP networking library (adapted for Nintendo Switch homebrew development)'
 arch=('any')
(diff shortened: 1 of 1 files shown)


LiquidFenrir:
Looking at the file linked by Subv for the latest release of libnx, they are not stubbed anymore for HOS 7.0.0 and above (something about an updated function), so this library should work?
Could we have this merged after testing please? Making a game with glfw and enet, would love to port it but enet isn't merged yet.

fincs:
Still needs libnx-side marshalling code to be written before recvmsg/sendmsg can work in homebrew, though.

switch/enet/PKGBUILD
```diff
@@ -0,0 +1,39 @@
+pkgname=switch-enet
+pkgver=r252.e2ef839
+pkgrel=1
+pkgdesc='ENet reliable UDP networking library (adapted for Nintendo Switch homebrew development)'
+arch=('any')
+url='https://github.com/lsalzman/enet'
+license=('MIT')
+options=(!strip libtool staticlibs)
+makedepends=('git' 'switch-pkg-config' 'devkitpro-pkgbuild-helpers')
+source=(${pkgname}::"git+https://github.com/lsalzman/enet.git#commit=e2ef83927d1626d6ee479742343a3011333fdfdd")
```
WinterMute, line 10: This can be done from a release tag now, just need to name the tarball we download, like this.

`${pkgname}-${pkgver}.tar.gz::${url}/archive/v${pkgver}.tar.gz`

switch/enet/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+md5sums=('SKIP')
+
+pkgver() {
```
WinterMute, line 13: This function not needed now.

switch/enet/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+    cd "$srcdir/${pkgname}"
+    printf 'r%s.%s' "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)"
+}
+
+build() {
+  cd ${pkgname}
+
+  source /opt/devkitpro/devkita64.sh
```
WinterMute, line 21: switchvars.sh pulls in devkita64.sh so we don't need to source both here.

switch/enet/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+    --disable-shared --enable-static
+
+  make
+}
+
+package() {
+  cd ${pkgname}
+
+  source /opt/devkitpro/devkita64.sh
+  source /opt/devkitpro/switchvars.sh
```
WinterMute, line 36: as above. Only one needed.

switch/enet/PKGBUILD
```diff
@@ -0,0 +1,39 @@
...
+pkgver=r252.e2ef839
+pkgrel=1
+pkgdesc='ENet reliable UDP networking library (adapted for Nintendo Switch homebrew development)'
+arch=('any')
+url='https://github.com/lsalzman/enet'
+license=('MIT')
+options=(!strip libtool staticlibs)
+makedepends=('git' 'switch-pkg-config' 'devkitpro-pkgbuild-helpers')
+source=(${pkgname}::"git+https://github.com/lsalzman/enet.git#commit=e2ef83927d1626d6ee479742343a3011333fdfdd")
+md5sums=('SKIP')
```
WinterMute, line 11: prefer to have sha256sum here.

switch/enet/PKGBUILD
```diff
@@ -0,0 +1,39 @@
+pkgname=switch-enet
+pkgver=r252.e2ef839
```
WinterMute, line 2: looks like we can use 1.3.15 directly now.

WinterMute (changes requested):
@yellows8 has implemented `sendmsg` and `recvmsg` in libnx now & it looks like we can use the latest upstream release. Hopefully everything should work with requested changes.

Subv:
I've updated the PR with the suggestions from @WinterMute. But please note that I do not currently have a way to test if the latest version of the library _actually_ works on a Switch.

WinterMute (approved):

PR #47, opened 2018-09-29
Update switch libmikmod driver

libjared:
This is an updated Switch driver for libmikmod. Improvements include:

- Double buffer output
- Updating the driver no longer blocks until the current buffer completes playing
- Fix mismatch of sample rates causing high pitch

An example usage can be viewed at [libjared/switch-music](https://github.com/libjared/switch-music).


libjared:
Good catch, I've made those changes.

PR #52, opened 2018-10-15
bzip.org mirror is gone, use Debian's mirror

davidgfnet:


davidgfnet:
The error message is quite interesting though. But that's another matter :D

PR #53, opened 2018-10-15
Remove hardcoded paths and use env vars

davidgfnet:
This would only require devkitpro to be exported externally, opinions?
Let me know what you think.

### DIFF
--- a/3ds-examples/PKGBUILD
+++ b/3ds-examples/PKGBUILD
@@ -17,6 +17,6 @@ groups=('3ds-dev')
 
 package() {
   rm $srcdir/${pkgname}-${pkgver}.tar.bz2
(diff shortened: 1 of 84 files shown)


fincs:
This doesn't work. The path to the DEVKITPRO directory will be fixed once the package is built. It is not possible to make pacman packages that install to a variable directory.

WinterMute:
We're standardising on /opt/devkitpro and the env vars at this point are mainly legacy.

Ideally we also want all toolchain users to have the same setup and to be using (dkp-)pacman to update rather than building things themselves. This helps a lot with support issues, especially when problems are caused by mismatched libraries and/or inappropriate modification of the toolchain directories.

If anything the PKGBUILDs which use $DEVKITPRO should use /opt/devkitpro instead.

PR #54, opened 2018-11-03
fix license

Oreo639:
I made a dumb mistake.


WinterMute:
Not a dumb mistake, xiph license is BSD, it's fine.

PR #55, opened 2018-11-08
Added switch-mbedtls

Black-Seraph:
The first half of getting SSL/TLS working on switch-curl, but given SSL/TLS is useful by itself as well, here's the PKGBUILD.


WinterMute:
See https://github.com/devkitPro/pacman-packages/tree/mbedtls

Still testing

PR #57, opened 2018-11-14
ppc: Add libgd and libmad

QuarkTheAwesome:
Heyo!
Not much to say here, just a couple libraries for ppc. libmad's patches are mostly lifted from Arch, giving it pkg-config support and a bugfix.
It's my first time working contributing with your build system, hopefully everything is OK with the pkgbuilds.
Thanks!
Ash

PR #58, opened 2018-11-14
ppc: Add libgd

QuarkTheAwesome:
Just a rehash of #57, sans-libmad.
Thanks again,
Ash

### DIFF
--- a/ppc/libgd/PKGBUILD
+++ b/ppc/libgd/PKGBUILD
@@ -1,0 +1,49 @@
+# Maintainer: Dave Murphy <davem@devkitpro.org>
+# Contributor: Ash Logan <ash [at] heyquark [dot] com>
+
+pkgname=ppc-libgd
+pkgver=2.2.5
+pkgrel=1
+pkgdesc="Library for the dynamic creation of images"
+url="https://libgd.github.io"
+license=("custom")
+arch=("any")
+depends=("devkitPPC" "ppc-libjpeg-turbo" "ppc-libpng" "ppc-freetype")
+makedepends=("ppc-pkg-config" "devkitpro-pkgbuild-helpers")
+options=("!buildflags" "staticlibs" "libtool" "!strip")
+source=("https://github.com/libgd/libgd/releases/download/gd-$pkgver/libgd-$pkgver.tar.xz"
+        "libgd-2.2.5-nosyslog.patch")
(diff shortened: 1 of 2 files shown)


ppc/libgd/PKGBUILD
```diff
@@ -0,0 +1,49 @@
...
+
+    source /opt/devkitpro/ppcvars.sh
+
+    make DESTDIR="$pkgdir" install
+    install -Dm644 COPYING "$pkgdir/$PORTLIBS_PREFIX/licenses/$pkgname/COPYING"
+
+    # These binaries all end up for PowerPC, so we shouldn't keep them
+    rm -rf "$pkgdir/$PORTLIBS_PREFIX/bin"
+    # Some build systems depend on doing -lgd
+    ln -s "libgd_static.a" "$pkgdir/$PORTLIBS_PREFIX/lib/libgd.a"
```
WinterMute: This will cause failures when the package is installed on windows, it'll need to be a copy

QuarkTheAwesome: Good catch, should be fixed in 52ce1da

WinterMute:
Hmm, the cmake build is failing quite badly on my machine.

I noticed there's autotools setup in there too & that works as well as not requiring the library to be copied/symlinked so I've switched to using that for the build. Attached failing build log.


[libgd.log](https://github.com/devkitPro/pacman-packages/files/2589442/libgd.log)

QuarkTheAwesome:
Seems that cmake isn't finding stdint.h, so it's assuming MSVC. Do we have the same makepkg helpers? I noticed that mine (built from this repo) are newer than the ones in dkp-libs (1.1.0-4 vs 1.0.6-1).
In any case, if autotools works, that's awesome. Not sure how I missed that the first time around :grin:
I know of at least one application that links with just `-lgd` and not `-lgd_static`, are you saying it would still work in that case? I guess I could just change the application, still possible to do code changes.

WinterMute:
the autotools setup produces libgd.a & not libgd_static.a. Not really sure why the cmake build would produce the latter tbh. Assuming MSVC on a linux box would be a pretty weird thing to do.

Looking through the cmake tests it's adding  -I/usr/include -I/usr/local/include to the search path which  is clearly a terrible idea for cross compiling.

QuarkTheAwesome:
Huh, that is really odd. In any case, glad to see it's working now.

PR #62, opened 2018-12-14
Add networking tcp/udp/rtmp and a few other small changes

jakibaki:
This was needed for the In-Home-Switching release, I can confirm that at the very least both tcp and udp work flawlessly with this.


switch/ffmpeg/PKGBUILD
```diff
@@ -25,16 +31,16 @@ build() {
...
     --extra-ldflags='-fPIE -L${PORTLIBS_PREFIX}/lib -L${DEVKITPRO}/libnx/lib' \
     --disable-runtime-cpudetect --disable-programs --disable-debug --disable-doc \
-    --disable-network --disable-hwaccels --disable-encoders \
+    --enable-network --disable-hwaccels --disable-encoders \
     --disable-avfilter --disable-avdevice --enable-swscale --enable-swresample \
     --enable-libass --enable-libfreetype --enable-libfribidi --enable-libtheora \
     --enable-filter='rotate,transpose' \
-    --disable-protocols --enable-protocol=file \
-    --disable-demuxers --enable-demuxer=h264,matroska,mov,ogg \
+    --disable-protocols --enable-protocol=file --enable-protocol=tcp --enable-protocol=udp --enable-protocol=rtmp \
```
carstene1ns: I guess these protocols could be concatenated.

switch/ffmpeg/fix_net.patch
```diff
@@ -0,0 +1,24 @@
+diff -ura pkg.orig/libavformat/os_support.c pkg.new/libavformat/os_support.c
```
fincs: Is there any way that libnx could be enhanced so that this patch is rendered unnecessary?

jakibaki: idk about the `HAVE_INET_ATON` patch but a lot of software assumes that `poll.h` is in `sys/poll.h` so I guess changing that would help

switch/ffmpeg/PKGBUILD
jakibaki: Fixed

switch/ffmpeg/fix_net.patch
jakibaki: >I'm not keen on -Ofast, it has interesting side effects. Have you tested the impact at all?

I haven't tested the impact of Ofast, just enabled it and noticed that none of the stuff I was using broke, I probably shouldn't have included it in the pr.

PR #63, opened 2018-12-19
Add GIFLIB

colin969:


WinterMute:
Can I ask why?

gif is a pretty old format, there are better ways to store images now.

It may also be worth looking for a library that supports multiple formats.

Mainly curious. It does seem like this library is well maintained so I'm not going to insist on having something different.

colin969:
SDL doesn't support GIF playback, and I'm pulling them from the internet - Giflib can be used as a relatively easy to way split the frames up into SDL Surfaces.

The alternative seemed to be converting them on demand with ffmpeg which doesn't sound great for performance.

ImageMagick is a more generic image manipulator as well, but it's a behemoth in comparison.

WinterMute:
merged manually with https://github.com/devkitPro/pacman-packages/commit/4ce0f48d54d7afd8c3f1c5558533f058509d70dc & package uploaded.

PR #64, opened 2018-12-22
Add compiler flags to switch.cmake

Armada651:
These compiler flags need to be specified for any library you're porting, might as well put them in the toolchain file.

### DIFF
--- a/pkgbuild-scripts/switch.cmake
+++ b/pkgbuild-scripts/switch.cmake
@@ -3,6 +3,9 @@ set(CMAKE_SYSTEM_PROCESSOR aarch64)
 
 set(SWITCH_LIBNX TRUE)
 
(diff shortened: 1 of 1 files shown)


pkgbuild-scripts/switch.cmake
```diff
@@ -3,6 +3,9 @@ set(CMAKE_SYSTEM_PROCESSOR aarch64)
 
 set(SWITCH_LIBNX TRUE)
 
+set(CMAKE_C_FLAGS "-march=armv8-a -mtune=cortex-a57 -mtp=soft -fPIE")
```
fincs: These two should probably match the values in switchvars.sh, namely:

```bash
export CFLAGS="-O2 -ffunction-sections -fdata-sections -march=armv8-a -mtune=cortex-a57 -mtp=soft -fPIC -ftls-model=local-exec"
export CXXFLAGS="${CFLAGS}"
```

PR #65, opened 2018-12-27
Switch: Add Python

linkmauve:
Most of this has been ported from the PyNX project’s Makefile.


WinterMute:
This whole thing is an unsalvageable bodge unfortunately.

Being honest I'm not entirely convinced that there's any sensible way to deal with cpython in this environment. MicroPython & CircuitPython seem to be a much better fit.

However it's done it really shouldn't be done like this.

PR #66, opened 2019-01-12
Add ogre package

compomega:
This library needs fnmatch built into newlib/libsysbase to link properly. This library needs zzlib lib to build.


switch/ogre3d/ogre-switch.patch
```diff
@@ -0,0 +1,690 @@
...
++
+ namespace Ogre {
+     //-----------------------------------------------------------------------
+     template<> Root* Singleton<Root>::msSingleton = 0;
+@@ -862,7 +866,11 @@ namespace Ogre {
+         // or break out by calling queueEndRendering()
+         mQueuedEnd = false;
+ 
++#ifdef __SWITCH__
++        while( !mQueuedEnd && appletMainLoop() )
```
fincs: This smells of hackjob. Surely there is another way that other platforms are using to handle message-pump code and detect window close events?

switch/ogre3d/ogre-switch.patch
```diff
@@ -0,0 +1,690 @@
...
+index 48fdc33..1a8dcbd 100644
+--- a/OgreMain/src/OgreSearchOps.cpp
++++ b/OgreMain/src/OgreSearchOps.cpp
+@@ -35,6 +35,12 @@ THE SOFTWARE.
+ #include <string.h>
+ #include <stdlib.h>
+ 
++#if OGRE_PLATFORM == OGRE_PLATFORM_SWITCH
++#if !defined(strdup)
++extern "C" { char* strdup(const char*); }
```
fincs: Why was it necessary to redefine strdup? We already have strdup...

switch/ogre3d/ogre-switch.patch
```diff
@@ -0,0 +1,690 @@
...
++        mVisible = false;
++        mClosed = true;
++    }
++
++    void SwitchEGLWindow::_notifySurfaceCreated(void* window, void* config)
++    {
++        // Set the new window.
++        mWindow = window;
++
++        nwindowSetDimensions((NWindow*)mWindow, mWidth, mHeight);
```
fincs: This is creating a 1920x1080 window (which is correct), but it is not handling docked/undocked state changes (and adjusting the crop/viewport accordingly). I suppose this will be handled in the future?

switch/ogre3d/ogre-switch.patch
```diff
@@ -0,0 +1,690 @@
...
++            EGL_DEPTH_SIZE,   24,
++            EGL_STENCIL_SIZE, 8,
++            EGL_NONE
++        };
++
++        eglChooseConfig(mEglDisplay, framebufferAttributeList, &mEglConfig,
++            1, &numConfigs);
++
++        mEglSurface = eglCreateWindowSurface(mEglDisplay, mEglConfig,
++            (NWindow*)mWindow, nullptr);
```
fincs: This `(NWindow*)` cast is superfluous because switch-mesa's EGL headers define the native window type to be `void*` as well.

switch/ogre3d/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+pkgname=switch-ogre
+pkgver=1.11.1
+pkgrel=1
+pkgdesc='scene-oriented, flexible 3D engine written in C++'
+arch=('any')
+url='https://www.ogre3d.org/'
+license=('MIT')
+options=(!strip libtool staticlibs)
+source=( "https://github.com/OGRECave/ogre/archive/v${pkgver}.tar.gz" "ogre-switch.patch")
+makedepends=('devkitA64' 'devkitpro-pkgbuild-helpers' 'switch-zlib'
```
fincs: zlib, zziplib and freetype should be in `depends`, not `makedepends`.

switch/ogre3d/ogre-switch.patch
compomega: I've tried to get the window size to switch when changing to and from docked mode and haven't been successful yet. Most attempts involve a crash unless I destroy the GL context, surface, display, etc. The problem one is the GL context and all the things lost when it goes away. Perhaps I'm missing some API that would help. So yeah, I gave up on it for the time being.

fincs: All you need to do is use nwindowSetCrop and glViewport. I'm not familiar at all with ogre, but presumably there's a way to update the viewport. Please refer to the dynamic_resolution example for more details.

switch/ogre3d/PKGBUILD
```diff
@@ -0,0 +1,52 @@
...
+
+  patch -Np1 -i "$srcdir"/ogre-switch.patch
+
+  FREETYPE_DIR=/opt/devkitpro/portlibs/switch cmake -G"Unix Makefiles" \
+    -DCMAKE_TOOLCHAIN_FILE=/opt/devkitpro/switch.cmake \
+    -DCMAKE_INSTALL_PREFIX=$PORTLIBS_PREFIX \
+    -DSWITCH=ON -DOGRE_STATIC=ON -DOGRE_BUILD_DEPENDENCIES=OFF \
+    -DOGRE_BUILD_RENDERSYSTEM_GL=OFF \
+    -DOGRE_BUILD_RENDERSYSTEM_GL3PLUS=ON \
+    -DZLIB_INCLUDE_DIR=/opt/devkitpro/portlibs/switch/include \
```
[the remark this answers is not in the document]
compomega: CMake is detecting zlib/zzip/freetype correctly now. It's probably a difference from the environment variables on my command line and after I started making the package build it. I'll remove those but the GL one is still not detecting right. There is this file:
```/opt/devkitpro/cmake/OpenGLConfig.cmake```

This file isn't used by any package and when I specify the module path to the file it doesn't actually detect:
```-DCMAKE_MODULE_PATH=/opt/devkitpro/cmake```

switch/ogre3d/ogre-switch.patch
WinterMute: you can use -std=gnu++11 or -D_GNU_SOURCE to enable strdup

compomega: Hmmm yeah the man page does actually say it's posix and not standard C. They probably shouldn't have used it or told CMake to use the GNU extensions (which they turned OFF at the top of the CMakeLists.txt). Oh well, time to wait for the whole library to slowly build again x.x

WinterMute:
Could you separate zziplib and submit that as a separate PR, please? Trying to keep things to 1 package per PR

compomega:
I updated this to the latest version and pulled in two more packages (switch-freeimage and switch-pugixml). I've setup a 2nd repository to use/share my packages here: https://github.com/compomega/devkitPro-repo

Of course I would rather they end up in this repository :smile: 

Was there any other problems I should address with the package?

PR #71, opened 2019-02-24
Add 3ds-flac

Oreo639:


WinterMute:
why --disable-cpplibs & --without-ogg?

Oreo639:
A while ago someone said they got it working and that they disabled ogg. I tested it rn and it works with it enabled.
As for cpplibs, that was disabled in switch-flac, so I kep it the same here. If there was any specific reason to disable it there and not there, I'll re-enable it.

3ds/flac/PKGBUILD
```diff
@@ -0,0 +1,47 @@
...
+)
+
+build() {
+  cd flac-$pkgver
+
+  source /opt/devkitpro/3dsvars.sh
+
+  patch -Np1 -i "$srcdir/flac-${pkgver}.patch"
+
+  CFLAGS="${CFLAGS} -D_3DS " \
```
WinterMute: -D_3DS is included in CPPFLAGS here - https://github.com/devkitPro/pacman-packages/blob/master/pkgbuild-scripts/3dsvars.sh#L8 can you check if it's working?

3ds/flac/PKGBUILD
```diff
@@ -0,0 +1,47 @@
...
+makedepends=('devkitpro-pkgbuild-helpers')
+groups=('3ds-portlibs')
+
+source=(
+  "https://ftp.osuosl.org/pub/xiph/releases/flac/flac-$pkgver.tar.xz"
+  "flac-${pkgver}.patch"
+)
+sha256sums=(
+  '91cfc3ed61dc40f47f050a109b08610667d73477af6ef36dcad31c31a4a8d53f'
+  'SKIP'
```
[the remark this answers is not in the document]
Oreo639: Woops.

3ds/flac/PKGBUILD
Oreo639: Yeah, it works.

PR #73, opened 2019-03-05
Add libsidplayfp

badda71:


3ds/libsidplayfp/PKGBUILD
```diff
@@ -0,0 +1,30 @@
...
+pkgname=3ds-libsidplayfp
+pkgver=1.8.8
+pkgrel=1
+pkgdesc="Library to play Commodore 64 music files (SID-files) derived from libsidplay2"
+arch=('any')
+url="https://sourceforge.net/projects/sidplay-residfp/"
+license=('GPLv2')
+options=(!strip libtool staticlibs)
+source=(
+	"https://netcologne.dl.sourceforge.net/project/sidplay-residfp/libsidplayfp/1.8/libsidplayfp-1.8.8.tar.gz"
```
WinterMute: This is a specific mirror. Replace the url here with `https://netcologne.dl.sourceforge.net/project/sidplay-residfp/libsidplayfp/${_basever}/libsidplayfp-${pkgver}.tar.gz`

3ds/libsidplayfp/PKGBUILD
```diff
@@ -0,0 +1,30 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+# Contributor: Sebastian Weber <me@badda.de>
+
+pkgname=3ds-libsidplayfp
+pkgver=1.8.8
```
WinterMute: change this to
```
_basever=1.8
_patchver=8
pkgver=${_basever}.${_patchver}
```
which will make the url change easier

badda71: ok

3ds/libsidplayfp/PKGBUILD
Oreo639: Just out of curiosity, what was the error?

badda71: I just tried again - the error is not occurring anymore, seems like sourceforge had only temporary issues:
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to downloads.sourceforge.net:443

WinterMute: It's just SourceForge mirror system messing up. It sometimes times out or returns a mirror that's currently down.

Oreo639:
Sorry for the late reply and thank you for your patience. (I meant to reply a while ago, but forgot about it)
I did some testing with sidplayfp and you were correct. It is way way way to slow on a 3ds. (at least in my testing it was)

badda71:
Is that apply for N3DS as well? I have a PR for the old libsidplay pending as well. The old libs perform better

Oreo639:
As for n3ds, I can't test that. As for the old sidplay, yes it works fine.

WinterMute:
Closing this one since the consensus is it's too slow on 3DS to be useful.

PR #77, opened 2019-03-14
Add libode to the switch ports

davidgfnet:
By default is build statically, with single precision, with ODE and libccd and without any thread support, should be enough for most homebrew.
Tested locally but with some custom hacks.

### DIFF
--- a/switch/ode/PKGBUILD
+++ b/switch/ode/PKGBUILD
@@ -1,0 +1,45 @@
+
+# Maintainer: davidgfnet <david@davidgf.net>
+pkgname=switch-ode
+pkgver=0.16
+pkgrel=1
+pkgdesc='High performance library for simulating rigid body dynamics'
+arch=('any')
+url='http://ode.org/'
+license=('BSD')
+options=(!strip libtool staticlibs)
+source=("https://bitbucket.org/odedevs/ode/downloads/ode-${pkgver}.tar.gz"
+        "threading.patch")
+makedepends=('devkitpro-pkgbuild-helpers')
+sha256sums=('4ba3b76f9c1314160de483b3db92b0569242a07452cbb25b368e75deb3cabf27'
+            '6f74df19aa2dfc9cd25fd8692e6a931594392a6afcdc768f2aca704224ddfe3d')
+groups=('switch-portlibs')
+
+prepare() {
+  cd ode-$pkgver
+  patch -Np1 < ../threading.patch
+}
+
+build() {
+  cd ode-$pkgver
+
+  source /opt/devkitpro/switchvars.sh
+
+  # Demos depend on GL and X11, disable since they are not useful to us.
+  # Disable threading for now, since pthreads is not usable out of the box for switch.
+  ./configure --prefix="${PORTLIBS_PREFIX}" --host=aarch64-none-elf \
+              --disable-demos --enable-static
+
+}
+
+package() {
+
+  cd ode-$pkgver
+
+  source /opt/devkitpro/switchvars.sh
+
+  make install DESTDIR="$pkgdir"
+
+}
+
+
--- a/switch/ode/threading.patch
+++ b/switch/ode/threading.patch
@@ -1,0 +1,10 @@
+--- a/ode/src/threading_pool_posix.cpp
++++ b/ode/src/threading_pool_posix.cpp
+@@ -371,7 +371,6 @@
+ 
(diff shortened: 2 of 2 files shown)


switch/libode/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+license=('BSD')
+options=(!strip libtool staticlibs)
+source=( "https://bitbucket.org/odedevs/ode/downloads/ode-${pkgver}.tar.gz" )
+makedepends=('devkitpro-pkgbuild-helpers')
+sha256sums=('4ba3b76f9c1314160de483b3db92b0569242a07452cbb25b368e75deb3cabf27')
+
+build() {
+  cd ode-$pkgver
+
+  source /opt/devkitpro/devkita64.sh
```
Oreo639: This is not needed. (already called in switchvars.sh)

switch/libode/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+  ./configure --prefix="${PORTLIBS_PREFIX}" --host=aarch64-none-elf \
+              --disable-threading-intf --disable-demos --enable-static
+
+}
+
+package() {
+
+  cd ode-$pkgver
+
+  source /opt/devkitpro/devkita64.sh
```
Oreo639: Same for here

switch/libode/PKGBUILD
```diff
@@ -0,0 +1,38 @@
...
+pkgrel=1
+pkgdesc='ODE (Open Dynamics Engine): real-time physics simulation featuring advanced collision with friction and joint types. Useful for games and simulation tools. (for Nintendo Switch homebrew development)'
+arch=('any')
+url='http://ode.org/'
+license=('BSD')
+options=(!strip libtool staticlibs)
+source=( "https://bitbucket.org/odedevs/ode/downloads/ode-${pkgver}.tar.gz" )
+makedepends=('devkitpro-pkgbuild-helpers')
+sha256sums=('4ba3b76f9c1314160de483b3db92b0569242a07452cbb25b368e75deb3cabf27')
+
```
Oreo639: Dont forget to add the switch-portlibs group.

davidgfnet:
Please take another look

switch/libode/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+# Maintainer: davidgfnet <david@davidgf.net>
+pkgname=switch-libode
```
WinterMute: make this switch-ode

switch/libode/PKGBUILD
```diff
@@ -0,0 +1,37 @@
...
+pkgver=0.16
+pkgrel=1
+pkgdesc='ODE (Open Dynamics Engine): real-time physics simulation featuring advanced collision with friction and joint types. Useful for games and simulation tools. (for Nintendo Switch homebrew development)'
```
WinterMute: Description here is a touch unwieldy - arch favours short descriptions that don't include package names, see https://wiki.archlinux.org/index.php/PKGBUILD#pkgdesc 

The [arch package here](https://www.archlinux.org/packages/community/x86_64/ode/) uses just "High performance library for simulating rigid body dynamics" which is probably fine. I know I've used "For X homebrew development" in quite a few other places but I think that's superfluous given that we already know what console we're targeting from the package prefix.

davidgfnet:
Done!

davidgfnet:
Friendly ping

WinterMute:
OK, can you see how this behaves with thread support enabled using latest devkitA64 & libnx?

davidgfnet:
It uses ptheads but AFAIK there's no pthread support in libnx right? There's some libnx-specific API that looks very similar to pthreads.

fincs:
There is pthread/std::thread support as of devkitA64 r13 and libnx 2.1.0.

davidgfnet:
Hey fincs, seems newlib in libnx does not provide pthread_attr_setinheritsched which seems being used by ODE

fincs:
We don't implement anything related to pthread priority settings, sorry. It might be considered in the future.

davidgfnet:
I guess we could indeed, but is it possible to implement that somehow?
I thought you were using newlib in libnx to cover most of these features.

davidgfnet:
Ok you guys wanna take a look at it? I just removed the scheduling thingy since I don't think we care for the switch port.

PR #85, opened 2019-04-28
Add 3ds-luajit

TSnake41:


WinterMute:
Installed files shouldn't be renamed or moved around, nor should the pkgconfig directory be deleted. The proper way to obtain cflags and libs for libraries like this is to use pkg-config rather than specifying them manually.

```
$ /opt/devkitpro/portlibs/3ds/bin/arm-none-eabi-pkg-config luajit --cflags
-I/opt/devkitpro/portlibs/3ds/include/luajit-2.0
$ /opt/devkitpro/portlibs/3ds/bin/arm-none-eabi-pkg-config luajit --libs
-L/opt/devkitpro/portlibs/3ds/lib -lluajit-5.1 -Wl,-E -lm -ldl
```

We can either supply a dummy libdl.a in a separate package or patch ./etc/luajit.pc  temporarily to avoid linker errors.

I also had some success building from top level dir rather than descending into src.

```
touch src/luajit src/jit/vmdef.lua
make TARGET_SYS=OTHER TARGET_T=libluajit.a  HOST_CC="gcc -m32"  XCFLAGS="-DLUAJIT_USE_SYSMALLOC" CFLAGS="" LDFLAGS="" LIBS="" HOST_CFLAGS="" HOST_LIBS="" HOST_LDFLAGS=""  CROSS=arm-none-eabi- PREFIX=${PORTLIBS_PREFIX} DESTDIR=`pwd`/test amalg install
```

Is there a good way to test the result?

TSnake41:
> We can either supply a dummy libdl.a in a separate package or patch ./etc/luajit.pc temporarily to avoid linker errors. -Wl,-E should be patched out anyway, we're not supplying dynamic libs.
A libdl library could be interesting if it dlfcn stuff works, e.g involves "modules" as structures, with minimal patching, we could build FFI with "libdl support".
Otherwise, we will have minimal FFI for embedded platforms (still capable though).

> Is there a good way to test the result?
Yes, I made a LuaJIT shell for 3DS : https://gist.github.com/TSnake41/cc723ffcfc386bbc9bf91cd1475e0836
After pressing (A), it shows a prompt where you can write Lua code.
e.g `print(jit.status())` checks JIT status (give true/false and JIT optimization features).

uyjulian:
I did it without moving files around:
https://github.com/uyjulian/LuaJIT/tree/3ds
`make install CROSS="arm-none-eabi-" HOST_CC="gcc -m32" TARGET_SYS=3DS DESTDIR=wherever`  
If you believe my changes are better, I can make a new PR with a patch

WinterMute:
@uyjulian This PR has already been updated to not move files around. I just haven't had time to test the library.

Your changes are intrusive and inappropriate, sorry.

uyjulian:
May I ask how this manages to get JIT working without mmap/etc implementation?

fincs:
It doesn't. However, hacking in a questionable implementation of mmap and pthread isn't a desirable solution either.

WinterMute:
One package and **only** one package per PR please. Separate out the stub libdl

TSnake41:
Making a cleanup, will reopen it later.

WinterMute:
protip: use `git pull --rebase` rather than just a pull when attempting to sync with master. This is one of the reasons why it's a good idea to work on a feature branch rather than directly on the master branch of a fork.

PR #100, opened 2019-07-05
Add libjansson to ppc

Crayon2000:
The _PKGBUILD_ file is based on the 3DS version but this one uses ppc and libjansson version 2.12. libjansson v2.11 is also used for the Switch.

I have kept 3 commits because I think it might be easier to review the changes this way. If accepted I propose to Squash all commits into one before merging.

### DIFF
--- a/ppc/jansson/PKGBUILD
+++ b/ppc/jansson/PKGBUILD
@@ -16,7 +16,6 @@ sha256sums=('5f8dec765048efac5d919aded51b26a32a05397ea207aa769ff6b53c7027d2c9')
 build() {
   cd jansson-$pkgver
 
-  source /opt/devkitpro/devkitppc.sh
   source /opt/devkitpro/ppcvars.sh
 
   ./configure --prefix="${PORTLIBS_PREFIX}" --host=powerpc-eabi \
@@ -27,7 +26,6 @@ build() {
 package() {
   cd jansson-$pkgver
 
-  source /opt/devkitpro/devkitppc.sh
   source /opt/devkitpro/ppcvars.sh
 
   make DESTDIR="$pkgdir" install
--- a/ppc/jansson/PKGBUILD
+++ b/ppc/jansson/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+# Contributor: Cpasjuste <cpasjuste@gmail.com>
+
+pkgname=3ds-jansson
+pkgver=2.11
+pkgrel=1
+pkgdesc='A C library for encoding, decoding and manipulating JSON data.'
+arch=('any')
+url='http://www.digip.org/jansson/'
+license=('custom')
+options=(!strip libtool staticlibs)
+makedepends=('devkitpro-pkgbuild-helpers')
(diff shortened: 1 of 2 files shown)


ppc/jansson/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+license=('custom')
+options=(!strip libtool staticlibs)
+makedepends=('devkitpro-pkgbuild-helpers')
+source=("http://www.digip.org/jansson/releases/jansson-$pkgver.tar.gz")
+sha256sums=('5f8dec765048efac5d919aded51b26a32a05397ea207aa769ff6b53c7027d2c9')
+
+build() {
+  cd jansson-$pkgver
+
+  source /opt/devkitpro/devkitppc.sh
```
WinterMute: remove source of devkitppc.sh, only ppcvars is required

ppc/jansson/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+
+  ./configure --prefix="${PORTLIBS_PREFIX}" --host=powerpc-eabi \
+    --disable-shared --enable-static
+  make
+}
+
+package() {
+  cd jansson-$pkgver
+
+  source /opt/devkitpro/devkitppc.sh
```
WinterMute: Same here, just ppcvars.sh

Crayon2000:
@WinterMute, changes were done for ppc.

Should the same changes be done for Switch and 3DS?

_3ds/jansson/PKGBUILD_:
```
  source /opt/devkitpro/devkitarm.sh
  source /opt/devkitpro/3dsvars.sh
```
_switch/jansson/PKGBUILD_:
```
  source /opt/devkitpro/devkita64.sh
  source /opt/devkitpro/switchvars.sh
```

Also, should the version be changed to jansson 2.12?

ppc/jansson/PKGBUILD
```diff
@@ -0,0 +1,36 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+# Contributor: Cpasjuste <cpasjuste@gmail.com>
```
WinterMute: Replace Contributor here with you.

Crayon2000:
@WinterMute, I did the change you asked.

WinterMute:
> @WinterMute, changes were done for ppc.
> 
> Should the same changes be done for Switch and 3DS?
> 

Eventually, yes. It doesn't really need done urgently since it has no effect on the packaging. I tend to sort these when there's an update to the library rather than just changing them all for the sake of it.

I'm also adding all of these libraries to a relevant portlibs group as they get updated. See https://github.com/devkitPro/pacman-packages/blob/master/ppc/libjpeg-turbo/PKGBUILD#L14 for instance. The groups are nds-portlibs, 3ds-portlibs, switch-portlibs & ppc-portlibs. Once I add libraries with platform specific changes then there'll probably be cube-portlibs and wii-portlibs as well.

> Also, should the version be changed to jansson 2.12?

Probably. If you're up for checking it builds and works OK on other platforms then please do feel free to make a PR for those.

Crayon2000:
@WinterMute, I have added the lib to the ppc-portlibs group.

WinterMute:
Good job, thanks for contributing.

PR #101, opened 2019-07-11
Add 3ds-libdl.

TSnake41:
As possibly needed with #85. This is a _extendable_ stub implementation of libdl.

It implements a "fake" libdl that can be needed for some libraries (like LuaJIT FFI).
You can (if needed) change the functions bound to the stub library using dl_setfn().

### DIFF
--- a/3ds/libdl/PKGBUILD
+++ b/3ds/libdl/PKGBUILD
@@ -1,0 +1,27 @@
+# Contributor: Teddy Astie <TSnake@Outlook.fr>
+# Maintainer:
+
+pkgname=3ds-libdl
+pkgver=1.0.0
+pkgrel=1
+pkgdesc='Extensible libdl library for projects that needs it.'
+arch=('any')
+options=(!strip)
+makedepends=('devkitpro-pkgbuild-helpers')
+source=('dlfcn.c' 'dlfcn.h')
+sha256sums=('9c84be0ef4365a3b2ed3948426ecfbdf8116611d63d5ec42d4af8a551c20ef72' '37c16bc645be27cad69904a53973a8862c37890f7327f6b26998794c62527336')
+groups=('3ds-portlibs')
+
+build() {
+  source /opt/devkitpro/3dsvars.sh
+  
+  arm-none-eabi-gcc $CFLAGS -c -o dlfcn.o dlfcn.c
+  arm-none-eabi-ar rcu libdl.a dlfcn.o
+}
+
+package() {
+  source /opt/devkitpro/3dsvars.sh
+  mkdir -p $pkgdir$PORTLIBS_PREFIX/lib/ $pkgdir$PORTLIBS_PREFIX/include
+  cp libdl.a $pkgdir$PORTLIBS_PREFIX/lib/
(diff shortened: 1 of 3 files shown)


3ds/libdl/PKGBUILD
```diff
@@ -0,0 +1,27 @@
...
+options=(!strip)
+makedepends=('devkitpro-pkgbuild-helpers')
+source=('dlfcn.c' 'dlfcn.h')
+sha256sums=('9c84be0ef4365a3b2ed3948426ecfbdf8116611d63d5ec42d4af8a551c20ef72' '37c16bc645be27cad69904a53973a8862c37890f7327f6b26998794c62527336')
+groups=('3ds-portlibs')
+
+build() {
+  source /opt/devkitpro/3dsvars.sh
+  
+  arm-none-eabi-gcc $CFLAGS -c -o dlfcn.o dlfcn.c
```
fincs: Should use `$(CC)` here

3ds/libdl/PKGBUILD
```diff
@@ -0,0 +1,27 @@
...
+  arm-none-eabi-ar rcu libdl.a dlfcn.o
```
fincs: Should use `$(AR)` here

3ds/libdl/PKGBUILD
```diff
@@ -0,0 +1,27 @@
...
+}
+
+package() {
+  source /opt/devkitpro/3dsvars.sh
+  mkdir -p $pkgdir$PORTLIBS_PREFIX/lib/ $pkgdir$PORTLIBS_PREFIX/include
```
fincs: Nitpick, but include is missing a trailing slash.

3ds/libdl/dlfcn.c
```diff
@@ -0,0 +1,36 @@
+#include "dlfcn.h"
+
+#include <stddef.h>
+
+static dl_open open = NULL;
```
fincs: Nitpick, but `= NULL` does nothing in this context and can be left out.
Also, you might want to rename these to something like `g_open` that do not have generic names that conflict with existing well known POSIX symbols; and yes, I'm aware that these are static vars, however it can be confusing to read.

3ds/libdl/dlfcn.c
```diff
@@ -0,0 +1,36 @@
...
+{
+  open = new_open;
+  error = new_error;
+  sym = new_sym;
+  close = new_close;
+}
+
+void *dlopen(const char *module, int flag)
+{
+  return open ? (*open)(module, flag) : NULL;
```
fincs: `(*fnptr)(params...)` can be simplified to just `fnptr(params...)`.

3ds/libdl/PKGBUILD
TSnake41: ${CC} is currently not defined in 3dsvars.sh, however, ${TOOL_PREFIX} is defined to `arm-none-eabi-` and is actually used to define ${AR}, so to get ${CC}, and can use ${TOOL_PREFIX}gcc.

PR #104, opened 2019-07-21
Add libxmp

knight-ryu12:
Adding libxmo from CMarsuoka

### DIFF
--- a/3ds/libxmp/PKGBUILD
+++ b/3ds/libxmp/PKGBUILD
@@ -1,0 +1,29 @@
+# Maintainer: Chromaryu <knight.ryu12@gmail.com>
+pkgname=3ds-libxmp-git
+pkgver=4.4.1
+pkgrel=1
+pkgdesc="Library that supports over 90 module formats"
+arch=('any')
+url="http://xmp.sourceforge.net/"
+license=('GPL')
+options=(!strip libtool staticlibs)
+source=("$pkgname"::"git+https://github.com/cmatsuoka/libxmp.git")
+sha256sums=('SKIP')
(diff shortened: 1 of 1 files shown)


WinterMute:
We only package stable releases where possible. We don't use git head and we require a checksum for downloaded source.

I'm also not entirely sure that libxmp performs well enough on 3DS - we had some issues at one point with speed and size and ended up using libxmp-lite instead.

It might be good to request a more recent stable release from upstream but we're trying to avoid rebuilding configury during packaging so we'd prefer not to be using incomplete (i.e with configure not already generated) source tarballs.

knight-ryu12:
good enough, I'll change it.
I made some test homebrew (3ds-tuneplayer), it worked out well.


2019年7月22日(月) 22:17 Dave Murphy <notifications@github.com>:

knight-ryu12:
I've changed its path to point to stable release of it.

asiekierka:
>I'm also not entirely sure that libxmp performs well enough on 3DS - we had some issues at one point with speed and size and ended up using libxmp-lite instead.

For whatever it's worth, we use it in production for MegaZeux, albeit trimmed down in terms of code size to only cover the formats we support. As far as I know, there are no performance complaints about this area of the engine.

knight-ryu12:
> we'd prefer not to be using incomplete (i.e with configure not already generated) source tarballs.

I mean, autoconf then ./configure is used commonly. 

> there are no performance complaints about this area of the engine.
Yeah. I've tested on both n3ds and o3ds, it worked out fine (but yeah slowdown on o3ds being weak, but it managed to play it back.)

PR #106, opened 2019-08-17
Update switch-freetype to v2.10.1.

DarkMatterCore:


DarkMatterCore:
Seems like I did an oopsie with the commit history. Closing this right away.

PR #107, opened 2019-08-17
Update switch-freetype to v2.10.1.

DarkMatterCore:


switch/freetype/PKGBUILD
```diff
@@ -1,10 +1,11 @@
 # Maintainer:  Dave Murphy <davem@devkitpro.org>
 # Contributor: cpasjuste <cpasjuste@gmail.com>
 # Contributor: carstene1ns <dev f4ke de>
+# Contributor: Pablo Curiel <pabloacurielz@gmail.com>
 
 pkgname=switch-freetype
-pkgver=2.9
-pkgrel=3
+pkgver=2.10.1
+pkgrel=4
```
WinterMute: pkgrel resets to 1 when package version is updated

DarkMatterCore: Fixed. Thanks for pointing that out.

PR #113, opened 2019-09-29
switch: add libntfs-3g

rhyskoedijk:
Provides libntfs-3g. 
To use this you need to supply your own `ntfs_device_operations` struct during init. This is currently not available out-of-the-box (using `libnx`) as the API is designed around low-level block storage.

````c
/**
 * struct ntfs_device_operations -
 *
 * The ntfs device operations defining all operations that can be performed on
 * the low level device described by an ntfs device structure.
 */
struct ntfs_device_operations {
	int (*open)(struct ntfs_device *dev, int flags);
	int (*close)(struct ntfs_device *dev);
	s64 (*seek)(struct ntfs_device *dev, s64 offset, int whence);
	s64 (*read)(struct ntfs_device *dev, void *buf, s64 count);
	s64 (*write)(struct ntfs_device *dev, const void *buf, s64 count);
	s64 (*pread)(struct ntfs_device *dev, void *buf, s64 count, s64 offset);
	s64 (*pwrite)(struct ntfs_device *dev, const void *buf, s64 count,
			s64 offset);
	int (*sync)(struct ntfs_device *dev);
	int (*stat)(struct ntfs_device *dev, struct stat *buf);
	int (*ioctl)(struct ntfs_device *dev, int request, void *argp);
};
````

There is an (old) example of an implementation for the Nintendo Wii here:
https://github.com/rhyskoedijk/libntfs-wii/blob/master/source/gekko_io.c

I am considering creating a new library (`libblockstorage`?) which would provide the necessary low-level access to (USB) block storage. With a library that can provide generic block storage, this and other filesystem libraries (ext2/3/4, fatfs, etc) could then be used.

Please let me know if this is heading down the right track, I'm not able to find much documentation on best practices for contributing, any guidance would be appreicated.


fincs:
Regular homebrew should *not* use or link against random libraries providing filesystem services.

The right way to lay this out in Horizon OS is to have some sort of custom sysmodule/service running in the background, which would provide access to an object following the standard IFilesystem interface, that can then be mounted using libnx's fsdev wrapper. Ideally, at some point Atmosphère would have a registry of custom filesystems, but nothing has been decided or planned yet.

Either way, manually parsing filesystems within final homebrew apps is definitely a no-go. I'm still leaving this PR open because I guess it will be necessary to implement said custom sysmodule/service.

rhyskoedijk:
Thanks. Is there a forum/discord/contact for Atmosphere to discuss if/how I could help with the above?

tiliarou:
@rhyskoedijk [https://reswitched.team/discord](https://reswitched.team/discord)

rhyskoedijk:
superseded by https://github.com/devkitPro/pacman-packages/pull/345

PR #116, opened 2019-10-19
switch: add libsmb2

rhyskoedijk:
Unstable port of `libsmb2` (built against HEAD). 
Not sure if this should wait until the next upstream release, but it is working fine for my needs as-is so I thought I'd share it. I would appreciate any feedback on the comments, not sure if they can be done in a better way.

Kerberos auth is not supported, but simple auth (username/password) works well.
Share enumeration/discovery works well too.

Basic usage:
````csharp
socketInitializeDefault();

ctx = smb2_init_context();
url = smb2_parse_url(ctx, "smb://[<domain;][<username>@]<host>[:<port>]/")
smb2_set_security_mode(ctx, SMB2_NEGOTIATE_SIGNING_ENABLED);
smb2_set_user(ctx, url->user); // if auth is required
smb2_set_password(ctx, "PASSWORD"); // if auth is required
smb2_connect_share(ctx, url->server, "SHARE_NAME", NULL);

// do the thing...

smb2_disconnect_share(ctx);
smb2_destroy_url(url);
smb2_destroy_context(ctx);

socketExit();
````

Better examples here:
https://github.com/sahlberg/libsmb2/tree/master/examples

### DIFF
--- a/switch/libsmb2/libsmb2-91d6c8a-fix_getaddrinfo_hints.patch
+++ b/switch/libsmb2/libsmb2-91d6c8a-fix_getaddrinfo_hints.patch
@@ -1,0 +1,20 @@
+diff --git a/lib/socket.c b/lib/socket.c
+index 978f06d..156fad5 100644
+--- a/lib/socket.c
++++ b/lib/socket.c
+@@ -701,9 +747,14 @@ smb2_connect_async(struct smb2_context *smb2, const char *server,
+         } else {
+                 port = "445";
+         }
++        
++        struct addrinfo hints;
++        memset(&hints, 0, sizeof hints);
++        hints.ai_family = AF_INET; // AF_INET or AF_INET6 to force version
++        hints.ai_socktype = SOCK_STREAM;
+ 
+         /* is it a hostname ? */
+-        if (getaddrinfo(host, port, NULL, &ai) != 0) {
++        if (getaddrinfo(host, port, &hints, &ai) != 0) {
+                 free(addr);
+                 smb2_set_error(smb2, "Invalid address:%s  "
+                                "Can not resolv into IPv4/v6.", server);
--- a/switch/libsmb2/libsmb2-91d6c8a-fix_makefile_libnx_include.patch
+++ b/switch/libsmb2/libsmb2-91d6c8a-fix_makefile_libnx_include.patch
@@ -1,0 +1,12 @@
+diff --git a/lib/Makefile.am b/lib/Makefile.am
+index 72a86ec..fe133eb 100644
+--- a/lib/Makefile.am
++++ b/lib/Makefile.am
+@@ -4,6 +4,7 @@ lib_LTLIBRARIES = libsmb2.la
+ 
+ libsmb2_la_CPPFLAGS = -I$(abs_top_srcdir)/include \
+ 		      -I$(abs_top_srcdir)/include/smb2 \
++			  -I$(DEVKITPRO)/libnx/include \
+ 		      "-D_U_=__attribute__((unused))"
+ 
+ libsmb2_la_SOURCES = \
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: switch/libsmb2/libsmb2-91d6c8a-fix_makefile_libnx_include.patch [hunk 1]
```diff
@@ -0,0 +1,12 @@
+diff --git a/lib/Makefile.am b/lib/Makefile.am
+index 72a86ec..fe133eb 100644
+--- a/lib/Makefile.am
++++ b/lib/Makefile.am
+@@ -4,6 +4,7 @@ lib_LTLIBRARIES = libsmb2.la
+ 
+ libsmb2_la_CPPFLAGS = -I$(abs_top_srcdir)/include \
+ 		      -I$(abs_top_srcdir)/include/smb2 \
++			  -I$(DEVKITPRO)/libnx/include \
```
rhyskoedijk: Not sure why this is needed, but the only way I could get it to work.
Build doesn't seem to pick up the `libnx` includes from the `-isystem` already defined in `switchvars.sh`.

In `configure.ac` some of the header checks are failing.
e.g. `AC_CHECK_HEADERS([arpa/inet.h])` despite that header actually existing, it reports it as missing for some reason.

switch/libsmb2/libsmb2-91d6c8a-fix_getaddrinfo_hints.patch
```diff
@@ -0,0 +1,20 @@
...
++++ b/lib/socket.c
+@@ -701,9 +747,14 @@ smb2_connect_async(struct smb2_context *smb2, const char *server,
+         } else {
+                 port = "445";
+         }
++        
++        struct addrinfo hints;
++        memset(&hints, 0, sizeof hints);
++        hints.ai_family = AF_INET; // AF_INET or AF_INET6 to force version
++        hints.ai_socktype = SOCK_STREAM;
```
rhyskoedijk: Was always throwing error 22 without this. Found an existing issue reporting the same problem, copy/pasted solution. Probably a better way to fix this though...

fincs: Fwiw getaddrinfo with hints=NULL is currently broken. However, a libnx branch I'm working on (https://github.com/switchbrew/libnx/tree/socket-cleanup) fixes this.

rhyskoedijk: thanks, I'll remove this patch once fixed in libnx then.

WinterMute:
readv and writev don't belong here, please remove that last merge.

rhyskoedijk:
> readv and writev don't belong here, please remove that last merge.

Is there somewhere in the framework where I can reference these? I wasn't able to find them last time I looked.

WinterMute:
> @WinterMute
> 
> > readv and writev don't belong here, please remove that last merge.
> 
> Is there somewhere in the framework where I can reference these? I wasn't able to find them last time I looked.

I don't really know what you mean by "framework" exactly but readv and writev aren't currently available in any library we provide but they don't belong in this library and shouldn't be provided here. That last merge you made will prevent this library ever being merged here.

WinterMute:
@rhyskoedijk Unless you remove the inappropriate code from this PR I'll have to reject and close this.

See https://github.com/devkitPro/newlib/issues/18

It's not possible to guarantee that these functions will work correctly but they should not be provided by libsmb2.

rhyskoedijk:
Sorry, been distracted lately.
Framework was maybe the wrong word. Basically I looked everywhere in devkita64 didn't find it and so I just put it inline to get my project to work at the time. I'm not an expert at any of this so pointers are appreciated. I'll remove readv/writev when I next get the chance.

Should I add a PR for sys/uio to newlib? or is it unlikely to be accepted given your comment regarding HOS!=POSIX?

PR #120, opened 2019-10-27
switch: add libprojectm

rhyskoedijk:
Cross-platform music visualization. Open-source and Milkdrop-compatible.
https://www.youtube.com/watch?v=2dSam8zwSFw

Native preset factory is disabled. Milkdrop presets ok. Textures ok. Fonts have issues (work in-progress).

**TODO:**
- [ ] Fix "undefined references" error during linking against `-lprojectm` with FTGL enabled
- [ ] Update `config.imp` paths to use `sdmc://`


proconsule:
last libprjectM can be compiled wihtout patch, i used and tested it on NXMilk

this is the PKGBUILD i am using

https://github.com/proconsule/NXMilk/blob/master/libs/projectM/PKGBUILD

PR #121, opened 2019-11-06
add MBEDTLS_CMAC_C

HookedBehemoth:
This is needed for running hactool on the switch so I'd appreciate it getting merged.


fincs:
Hmm, why do you need to run hactool on the Switch? I thought that was a PC tool?

HookedBehemoth:
turned it into a lib that can easily be deployed on the switch but it won't link without cipher_cmac

fincs:
I mean, what's exactly that library going to be used for?

HookedBehemoth:
I'll use it to validate nca's on console but having a lib opens it to anyone to do anything hactool can do.

HookedBehemoth:
Only extracting the exact parts I need would make this unmaintainable.

PR #127, opened 2020-02-14
switch: add libarchive

krruzic:
I just used the patch from the 3ds version, not sure if that was necessary but I'm lazy. Compiled fine but didn't test.

### DIFF
--- a/switch/libarchive/PKGBUILD
+++ b/switch/libarchive/PKGBUILD
@@ -1,0 +1,53 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+
+pkgname=switch-libarchive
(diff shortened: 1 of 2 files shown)


WinterMute:
Genuine question ... given that you didn't test this then is it unreasonable to think you don't have a need or use for this? What then prompted you to produce a package for it?

krruzic:
Sorry it's being used by nx-shell so I'm reasonably confident it works. I made this for my personal use and opened a PR maybe a bit ahead of time. I'm writing code that uses it, there's code out there using it. I'll compile nx-shell against this packaged version to test it out.

fincs:
Superseded by #273.

Issue #130, opened 2020-03-28
3ds-curl not working for ftp URLs


badda71:
The current version of 3ds-curl will always exit with "select/poll error" when trying to download ftp urls. This can be fixed by removing "#define HAVE_POLL_FINE 1" in file lib/curl_config.h, line 540 after executing ./configure and before executing make.
Seems like the poll() implementation is buggy on the 3DS and curl should rather rely on select().

fincs:
On 3DS, we implement select with poll, so switching that up would accomplish nothing (if all, it would make it worse).

badda71:
Ok, so the define HAVE_POLL_FINE breaks something else on the 3DS. Anyways, getting rid of this define fixes curl for ftp URLs

badda71:
Just opened a PR: https://github.com/devkitPro/pacman-packages/pull/131

WinterMute:
This is working with latest libs.

PR #131, opened 2020-03-29
undefine HAVE_POLL_FINE in select.c

badda71:
This will make libcurl work with ftp urls


fincs:
I'm afraid this PR is not acceptable. I already explained why in the other thread: https://github.com/devkitPro/pacman-packages/issues/130#issuecomment-605533304

badda71:
So you would rather have a broken libcurl than include the #undef?

fincs:
I would rather have a broken libcurl than a poorly understood bodge.

PR #135, opened 2020-04-14
3ds: add luajit

TSnake41:
This is a cleaned up version of https://github.com/devkitPro/pacman-packages/pull/85 that got too complicated.

I also added few notable changes from original pull-request : 
 - there is no longer need of libdl (that required a slight source code hack while building) so I added a luajit.pc patch to remove the link to libdl (see luajit.pc.patch), LuaJIT will use default "console" behavior and throw and error when attempting to use ffi.load and few other functions (which is the expected behavior).
    - this will prevent potential breaking if some change in the source code affect this hack
    - as it is recognized as a "console" platform, libdl is not used

notes :
JIT is working with various optimization (ARMv6, VFPv2, and standard JIT optimizations) but assumes that XN-bit is not set on malloc-ed memory (this has to be verified).

In case of issues with JIT we can still set `-DLUAJIT_DISABLE_JIT` to disable JIT and still have FFI plus a significant performance boost compared to original Lua 5.1. Note that this setting is required for a LuaJIT Switch port.

Used in [LovePotion LuaJIT port](https://github.com/TSnake41/LovePotion/tree/luajit-new)

### DIFF
--- a/3ds/luajit/PKGBUILD
+++ b/3ds/luajit/PKGBUILD
@@ -9,16 +9,16 @@ url='http://luajit.org/'
 license=('MIT')
 options=(!strip)
 makedepends=('3ds-pkg-config' 'devkitpro-pkgbuild-helpers')
-source=("https://luajit.org/download/LuaJIT-$pkgver.tar.gz" "luajit.pc.patch")
-sha256sums=('874b1f8297c697821f561f9b73b57ffd419ed8f4278c82e05b48806d30c1e979'
-            'b993f4bfb7dab888f909f6973996783d481d4cf767eca1d8a68df45202fcab9e')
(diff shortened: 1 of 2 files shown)


WinterMute:
Would be simpler to just sed the .pc file in place like we've done in some of the other PKGBUILDS. See https://github.com/devkitPro/pacman-packages/blob/master/3ds/SDL_mixer/PKGBUILD#L33 for instance. Saves adding a patch file & another checksum for such a small change.

The `-Wl,-E` should also be removed.

TSnake41:
I am not a master of sed, but still made something working.

uyjulian:
`echo -n (string)` is non-POSIX. I would recommend you use `printf '%s' (string)` instead.

PR #136, opened 2020-04-15
Update 3ds-mbedtls to 2.16.6.

rinsuki:


carstene1ns:
Can you fix `description` and `url` too? :+1:

mtheall:
Can you do the switch port as well?

rinsuki:
@carstene1ns Fixed 👍

@mtheall sorry, I can't test on real machine because I don't have Switch CFW/Homebrew environment. If still good, I make a patch.

carstene1ns:
Should be straightforward, the switch patch is small enough.

rinsuki:
Created #137

WinterMute:
And while you're tidying up in here may as well replace the config.h part of the patch with the script used to enable/disable features like we did for switch.

See https://github.com/devkitPro/pacman-packages/blob/master/switch/mbedtls/PKGBUILD#L26-L30

rinsuki:
@WinterMute I did it! so can you review?

WinterMute:
Nice job, thanks for doing this.

PR #137, opened 2020-04-15
Update switch-mbedtls to 2.16.6.

rinsuki:
NOTE: I didn't confirm this actually work, but it probably does. At least the build was successful.


mtheall:
This appears to work for me.

PR #139, opened 2020-04-16
switch: add freeimage

compomega:
Because I know you will ask, I updated libpng to the latest to match the version in this repository. I did this because I saw PNG images loading wrong (looked garbled) in ogre and updating it to match fixed the problem. I was also linking to freetype which needs libpng. Perhaps the linker order would have fixed it but I didn't want to break freetype. Wish this library didn't embed other libraries like that. It works fine for shared objects but not so well for static libraries.


WinterMute:
Any package PR'd with a custom Makefile instead of using the original package build system will be rejected.  Embedding libraries that are already packaged will also be rejected since this will cause dependency issues and duplicate symbols.

PR #140, opened 2020-04-16
switch: add physfs

compomega:


switch/physfs/PKGBUILD
```diff
@@ -0,0 +1,46 @@
+# Maintainer: COMP Omega <compomega@tutanota.com>
+pkgname=switch-physfs
+pkgver=3.0.2
+pkgrel=2
+pkgdesc='scene-oriented, flexible 3D engine written in C++'
```
RSDuck, line 5: I'm pretty sure that's left from another package 😄

Stary2001:
Hi,
I've cleaned up this patch a bit as we have pthreads now, so don't need to override any of the mutex stuff. I also pulled the switch specific stuff into its own platform so we don't have to patch the unix one.
see https://github.com/Stary2001/pacman-packages/tree/switch-physfs/switch/physfs - I can make a new PR if needed.

carstene1ns:
@Stary2001: Should be able to move the patches from posix platform as well, my intention was to not be repetitive so much.

WinterMute:
probably worth just opening a new PR with the cleaner patches.

compomega:
Thanks for the update to the patch! I tried it out with the application I originally needed it for and it's still working for me.

You can take ownership of the library port and open a new PR or if you don't want to bother I can pull in your new patch to my PR. Either way works for me.

Stary2001:
I'm happy to take over physfs - I've made a new PR at #148 .

compomega:
Thanks for taking it over!

PR #142, opened 2020-04-20
switch-ffmpeg: update dependencies to remove switch-libopus, switch-libtheora, switch-libvpx

uyjulian:
`switch-libopus`, `switch-libtheora`, and `switch-libvpx` can be removed, as `ffmpeg` has builtin decoders for these formats.

The builtin codecs are more optimized than `libopus`, `libtheora`, and `libvpx`.

### DIFF
--- a/switch/ffmpeg/PKGBUILD
+++ b/switch/ffmpeg/PKGBUILD
@@ -5,7 +5,7 @@
 
 pkgname=switch-ffmpeg
 pkgver=4.2.2
-pkgrel=1
+pkgrel=2
 pkgdesc='ffmpeg port (for Nintendo Switch homebrew development)'
 arch=('any')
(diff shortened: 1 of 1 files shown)


carstene1ns:
It might be true for decoding, but you can also encode with ffmpeg. No idea if one would ever want to convert videos or such on switch, but without these libs, the supported formats decrease.

For example the ffmpeg docs say:
>opus: Opus encoder.
>
>This is a native FFmpeg encoder for the Opus format. Currently its in development and only implements the CELT part of the codec. Its quality is usually worse and at best is equal to the libopus encoder.

uyjulian:
Encoding is already disabled on the configure command line.

WinterMute:
If there's a need for encoding on Switch we can cross that bridge when we come to it.

PR #144, opened 2020-05-09
switch-curl: libnx tls backend and system proxy support.

yellows8:
This replaces the mbedtls tls backend with the newly implemented libnx tls backend. Also adds support for using the system proxy from nifmGetCurrentNetworkProfile.

This uses functionality currently only in libnx-git, so this is blocking on the next libnx stable release.


MstrVLT:
#153

PR #145, opened 2020-05-12
switch: add libffi

friedkeenan:
Needed it for some python stuff, the `_ctypes` module specifically.

It says `libtool: warning: remember to run 'libtool --finish /opt/devkitpro/portlibs/switch/lib'` near the end of the `make install`, and this is the first time I've written a pkgbuild so I'm not sure if I'm supposed to worry about that and if so what I'm supposed to do.

### DIFF
--- a/switch/libffi/PKGBUILD
+++ b/switch/libffi/PKGBUILD
@@ -1,0 +1,41 @@
+# Maintainer: friedkeenan <friedkeenan@protonmail.com>
+
+pkgname=switch-libffi
+pkgver=3.3
+pkgrel=1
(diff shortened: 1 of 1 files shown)


uyjulian:
Is there a reason to have available `_ctypes` module when dynamic libraries are not supported when using `libnx`?

friedkeenan:
Yes, I want to be able to use the `ctypes` module for the simple data types (`c_uint32`, `c_char`, etc.) and for structures and stuff, and that relies on the `_ctypes` module

WinterMute:
No.

friedkeenan:
Okie dokie then

PR #148, opened 2020-05-25
switch: add physfs

Stary2001:
Cleaned up PhysFS patch using pthreads support. Continuation of #140

### DIFF
--- a/switch/physfs/PKGBUILD
+++ b/switch/physfs/PKGBUILD
@@ -12,7 +12,7 @@ source=( "https://icculus.org/physfs/downloads/physfs-3.0.2.tar.bz2"
 makedepends=('devkitA64' 'devkitpro-pkgbuild-helpers')
 sha256sums=(
   '304df76206d633df5360e738b138c94e82ccf086e50ba84f456d3f8432f9f863'
-  'SKIP'
+  'a30e2ccc16dd4774e78e95e2cc9e1fbb437bf0aa8113241f6cd054d0860c03ab'
   )
 groups=('switch-portlibs')
 build() {
--- a/switch/physfs/PKGBUILD
+++ b/switch/physfs/PKGBUILD
@@ -1,0 +1,41 @@
+# Maintainer: Ezekiel Bethel <zek@9net.org>
+pkgname=switch-physfs
+pkgver=3.0.2
+pkgrel=1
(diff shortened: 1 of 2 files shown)


switch/physfs/physfs.patch
```diff
@@ -0,0 +1,191 @@
...
++		/* nx-hbmenu should give us the full path of the application, this may
++		 *  reside in a subfolder. Higher level code will strip away the name
++		 *  and extension.
++		 */
++		return NULL;
++	}
++
++	if (!retval)
++		/* Last resort: use `/switch` directory. */
++		retval = __PHYSFS_strdup("sdmc:/switch/");
```
WinterMute, line 163: Change this to just "/switch/" - avoiding explicit device names will allow code to run from different devices in the future. Is the trailing slash needed?

switch/physfs/physfs.patch
```diff
@@ -0,0 +1,191 @@
...
++    return __PHYSFS_switchCalcUserDir();
++#endif
+     char *retval = NULL;
+     char *envr = getenv("HOME");
+ 
+@@ -299,7 +302,14 @@
+ int __PHYSFS_platformStat(const char *fname, PHYSFS_Stat *st, const int follow)
+ {
+     struct stat statbuf;
++#ifdef PHYSFS_PLATFORM_SWITCH
```
WinterMute, line 43: This part is no longer needed since we implemented lstat via stat

switch/physfs/PKGBUILD
```diff
@@ -0,0 +1,41 @@
...
+arch=('any')
+url='https://icculus.org/physfs/'
+license=('zlib')
+options=(!strip libtool staticlibs)
+source=( "https://icculus.org/physfs/downloads/physfs-3.0.2.tar.bz2"
+         "physfs.patch")
+makedepends=('devkitA64' 'devkitpro-pkgbuild-helpers')
+sha256sums=(
+  '304df76206d633df5360e738b138c94e82ccf086e50ba84f456d3f8432f9f863'
+  'SKIP'
```
WinterMute, line 15: provide sha256sum value here please.

switch/physfs/physfs.patch
carstene1ns: Yes, trailing slash is added everywhere else as well.
> Platform layer is required to append a dirsep

PR #149, opened 2020-06-12
pkgbuild-scripts: Add wiivars.sh and wiiuvars.sh to the list of devkitPPC files to install

sergiou87:
Not sure if this is the right way to submit this change 😂  I'm trying to make SDL for Wii install using the official build tools in case we can make it a pacman package (https://github.com/dborth/sdl-wii/issues/54) and noticed that vars file was missing.

Also I haven't changed the package version or `rel` number, happy to change those too if needed 😄 

EDIT: I just added `wiiuvars.sh` as requested 👍


WinterMute:
can you add wiiuvars.sh as well while you're in there please?

sergiou87:
Done!

PR #150, opened 2020-06-14
Add SDL and SDL_mixer packages for Gamecube/Wii

sergiou87:
Disclaimer: This PR is just a draft/placeholder for now to give some visibility to this work and make sure I'm doing the right thing here.

This is still WIP and would introduce the following changes:
 - Add SDL package for Wii and Gamecube. The patch would need to be generated from https://github.com/devkitPro/SDL/tree/wii-sdl1.2 when all the changes are ready.
 - Add SDL_mixer package for Wii and Gamecube. The patch would need to be generated from https://github.com/devkitPro/SDL_mixer/tree/sdl_mixer-1.2-big-endian


WinterMute:
We prefer one PR per package rather than adding multiple packages in one go like this.

I'm happy to give you access to https://github.com/devkitPro/SDL so we can keep all the ports in one place as much as possible.

sergiou87:
Sure! The intention is to give the visibility of all I've been doing and track all the changes, in case I'm doing something wrong or disappear for work or whatever 😂  Happy to submit one PR per package once everything is more stable 😄 

Thanks for the access to the SDL repo. I guess you also need to fork SDL-mirror/SDL_mixer to devkitPro so I can push my endianness fixes there too.

sergiou87:
Actually, I can submit a PR for ppc-vorbis anyway (and the 3ds-sdl_mixer description change? 😅) if you agree with using it for SDL_mixer on Wii/GC, instead of Tremor.

WinterMute:
Oops, sorry, I already added ppc-libvorbis earlier when I was looking at the wiiu link error. I'm fine with using it for Wii/Gamecube SDL.

The SDL packages should go under wii and gamecube with appropriate package renaming. PR for 3ds-sdl_mixer would be much appreciated.

I added a SDL_mixer fork & added you there now.

Thanks so much for taking this on. It's very much appreciated.

sergiou87:
Already taken care of, closing… thanks for the help! 😄

WinterMute:
Really. Thank you. You've been a massive help in getting these together.

PR #154, opened 2020-06-21
Add packages to support GP2X

Orkie:
This package adds contains crtls for GP2X and gp2x_rules, for use with devkitARM, within the gp2x-dev group.


WinterMute:
Could you split these to separate PRs per package please?

I'm also going to separate libfat packages rather than trying to keep them all in one PKGBUILD so would prefer that as a gp2x/libfat/PKGBUILD that only builds gp2x variant. I keep tweaking libfat for one platform which then doesn't need other paltforms rebuilt and it all gets a little weird.

Orkie:
Will do

Orkie:
Closing to move to separate PRs

PR #169, opened 2020-07-13
add msgpack-c

HookedBehemoth:
Depends on #167


switch/msgpack/switch-msgpack.patch
```diff
@@ -0,0 +1,13 @@
+diff -NBaur msgpack-3.3.0.switch/CMakeLists.txt msgpack-3.3.0/CMakeLists.txt
+--- msgpack-3.3.0/CMakeLists.txt	2018-12-02 13:20:52.000000000 +0000
++++ msgpack-3.3.0.switch/CMakeLists.txt	2018-12-23 18:49:07.178930622 +0000
+@@ -5,6 +5,8 @@
+     CMAKE_POLICY(SET CMP0054 NEW)
+ ENDIF ()
+ 
++set(CMAKE_C_COMPILER_WORKS TRUE)
```
WinterMute, line 8: Setting these variables manually shouldn't be necessary, what error occurs without them?

WinterMute:
The only dependency I can see on #167 is in the example code & the example building is disabled. Is there a reason you think there's still a dependency here?

switch/msgpack/switch-msgpack.patch
HookedBehemoth: When trying to compile the cmake test program it fails as _start isn't defined without linking against libnx

HookedBehemoth:
Oh you are correct.
I closed the cJSON PR.

PR #172, opened 2020-08-05
libogg, libvorbis: update to 1.3.4 and 1.3.7

asiekierka:


WinterMute:
Single PR per package please. Multiple platforms in one PR for a single package are fine provided that they're all the same version.

PR #174, opened 2020-08-09
mbedtls: implement timing

Stary2001:
I didn't test an actual application using the timing API yet (specifically Godot's DTLS multiplayer peer), but I have tested the implementation of mbedtls_set_alarm. mbedtls_timing_get_delay/mbedtls_timing_set_delay use gettimeofday, which should function properly.


switch/mbedtls/switch-mbedtls.patch
```diff
@@ -51,3 +51,104 @@ diff -NBaur mbedtls-2.14.1/library/net_sockets.c mbedtls-2.14.1.switch/library/n
...
++#ifdef __SWITCH__
++static bool timer_thread_created = false;
++static unsigned int alarmUs = 0;
++static Thread timerThread;
++static alignas(0x1000) char timerThreadStack[0x1000];
++
++static void timer_thread_func(void *param)
++{
++    (void) param;
++    usleep(alarmUs);
```
fincs, line 107: If we're going to use native apis, then `svcSleepThread((s64)alarmUs * 1000L);` would be a better thing to do here.

switch/mbedtls/switch-mbedtls.patch
```diff
@@ -51,3 +51,104 @@ diff -NBaur mbedtls-2.14.1/library/net_sockets.c mbedtls-2.14.1.switch/library/n
...
++        mbedtls_timing_alarmed = 1;
++        return;
++    }
++
++    mbedtls_timing_alarmed = 0;
++    alarmUs = seconds * 1000000;
++
++    if(timer_thread_created)
++    {
++        threadWaitForExit(&timerThread);
```
fincs, line 133: Is this the desired behavior? This logic is saying: if there's an ongoing timer, this is going to wait for *that* timer to finish first, then afterwards set the new one.

switch/mbedtls/switch-mbedtls.patch
```diff
@@ -51,3 +51,104 @@ diff -NBaur mbedtls-2.14.1/library/net_sockets.c mbedtls-2.14.1.switch/library/n
...
++        threadClose(&timerThread);
++    }
++
++    timer_thread_created = true;
++
++    threadCreate(&timerThread, timer_thread_func, NULL, timerThreadStack, 0x1000, 0x3b, 0);
```
fincs, line 139: core_id parameter could be `-2` instead of `0` in order to not pin the thread to core0.

switch/mbedtls/switch-mbedtls.patch
```diff
@@ -51,3 +51,104 @@ diff -NBaur mbedtls-2.14.1/library/net_sockets.c mbedtls-2.14.1.switch/library/n
...
++
+ #if !defined(HAVE_HARDCLOCK)
+ 
+ #define HAVE_HARDCLOCK
+@@ -313,6 +327,51 @@
+     }
+ }
+ 
++#ifdef __SWITCH__
++static bool timer_thread_created = false;
```
fincs, line 99: I think this variable should be switched over to atomic load/store.

switch/mbedtls/switch-mbedtls.patch
```diff
@@ -51,3 +51,104 @@ diff -NBaur mbedtls-2.14.1/library/net_sockets.c mbedtls-2.14.1.switch/library/n
...
++        threadWaitForExit(&timerThread);
++        threadClose(&timerThread);
++    }
++}
++
++void mbedtls_set_alarm( int seconds )
++{
++    if(seconds == 0)
++    {
++        mbedtls_timing_alarmed = 1;
```
fincs, line 124: I think this variable should be switched over to atomic load/store. Currently this is highly UB on Arm processors due to out-of-order memory accesses 🙈

switch/mbedtls/switch-mbedtls.patch
Stary2001: I think the "correct" behaviour here is unspecified - on Windows it'll leave the current timer running _and_ start a new one(!) so whichever is first wins, and on Linux with alarm() it'll cancel the current timer first. I agree this is probably not what we want though, but I'm unsure how to implement it without just keeping a list of thread handles...really nobody should be using this API because it sucks anyway.

switch/mbedtls/switch-mbedtls.patch
Stary2001: I switched to storing alarmSecs and used alarmSecs * 1e9L as we get given seconds initially. No need to multiply it to us, store it, then multiply again to ns.

PR #181, opened 2020-08-13
Armv4t libogg

Orkie:


asiekierka:
Might be a good idea to update to 1.3.4, as per https://github.com/devkitPro/pacman-packages/pull/201

Orkie:
> Might be a good idea to update to 1.3.4, as per #201

Yes, done!

PR #204, opened 2020-10-25
switch: add GLEW

Exvat:
This should make it easier to port projects that rely on GLEW.

Some patches were necessary because GLEW assumes that the system allows you to like directly to all OpenGL 1.1 functions. This is not true for the the implementation in devkitpro which requires that all functions are loaded from EGL. To solve this the patch enables GLEW to also wrangle OpenGL 1.1.


halotroop2288:
How did you submit this from an "unknown repository"?
image

WinterMute:
> How did you submit this from an "unknown repository"? image

The account that submitted the PR has been removed. Please don't spam our PRs. We get quite enough email from this stuff as it is without necrobumping and marking PRs as "approved".

PR #206, opened 2020-12-23
switch-mesa: update to 20.1.10

deltabeard:
Update switch-mesa from 20.1.0-rc3 to 20.1.10, which is the latest patch of Mesa in the 20.1 series at the time of writing.

This draft pull request requires further testing.
The PKGBUILD file was modified to remove OpenGLES support (leaving on OpenGL). This may not be acceptable.

Signed-off-by: Mahyar Koshkouei <mk@deltabeard.com>


deltabeard:
switch-sdl2 used OpenGLES2, so I will re-add this.

WinterMute:
switch mesa is not maintained here. Patches are generated from https://github.com/devkitPro/mesa and the repo there will be rebased against newer upstream when necessary.

PR #212, opened 2021-02-03
switch: add libsmm

Witherking25:


Witherking25:
This is a library for making simple mod manager plugins

PR #213, opened 2021-02-17
New package: libwebsockets

ConnorRigby:
This adds support for [libwebsockets](https://libwebsockets.org/)
the port is pretty simple, mostly the same as the `freertos`  platform used by esp32

### DIFF
--- a/switch/libwebsockets/0001-Add-platform-Nintendo-Switch.patch
+++ b/switch/libwebsockets/0001-Add-platform-Nintendo-Switch.patch
@@ -1,0 +1,1112 @@
+From 1fa3221a01402910b45bcd8a60f2fba5d02dcc40 Mon Sep 17 00:00:00 2001
+From: Connor Rigby <connorr@hey.com>
+Date: Wed, 17 Feb 2021 06:39:38 -0800
+Subject: [PATCH] Add platform: Nintendo Switch
+
+---
+ CMakeLists.txt                             |   1 +
+ cmake/lws_config.h.in                      |   1 +
+ include/libwebsockets.h                    |   3 +-
+ include/libwebsockets/lws-network-helper.h |   4 +
+ lib/CMakeLists.txt                         |   2 +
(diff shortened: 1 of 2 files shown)


switch/libwebsockets/0001-Add-platform-Nintendo-Switch.patch
```diff
@@ -0,0 +1,1112 @@
...
++*/
++
++#include <unistd.h>
++#include <sys/types.h>
++#include <netdb.h>
++#include <netinet/in.h>
++#include <arpa/inet.h>
++#include <fcntl.h>
++#include <switch/kernel/mutex.h>
++#include <switch/services/bsd.h>
```
WinterMute, line 1052: don't use this bsd header. We translate between newlib fds & horizon fds internally. Use standard <sys/socket.h>

switch/libwebsockets/0001-Add-platform-Nintendo-Switch.patch
```diff
@@ -0,0 +1,1112 @@
...
++#define LWS_ENOTCONN ENOTCONN
++#define LWS_EWOULDBLOCK EWOULDBLOCK
++#define LWS_EADDRINUSE EADDRINUSE
++#define lws_set_blocking_send(wsi)
++#define LWS_SOCK_INVALID (-1)
++
++#define compatible_close(x) close(x)
++#define lws_plat_socket_offset() 0
++#define wsi_from_fd(A,B)  A->lws_lookup[B - lws_plat_socket_offset()]
++#define shutdown(sockfd, how) bsdShutdown(sockfd, how)
```
WinterMute, line 1071: shutdown is found in <sys/socket.h> don't use the bsd functions directly.

WinterMute (changes requested):

ConnorRigby:
Thanks for the review @WinterMute!
I just moved so it might take me a few more days to get to this, but i'll certainly get too it this week.

PR #214, opened 2021-02-21
Wii: enable ModPlug support for SDL_mixer

sergiou87:
While working on sergiou87/open-supaplex#13 I noticed SDL_mixer for Wii wasn't built with mod(plug) music support. This PR enables that.

However, before merging this, there are additional changes in devkitPPC and maybe somewhere else to fix:
1. SDL_mixer references `#include "modplug.h"` but that header is located in `libmodplug/modplug.h`
2. There is also a broken reference in `/opt/devkitpro/portlibs/ppc/lib/libmodplug.la`, where `dependency_libs` uses the wrong path.

### DIFF
--- a/wii/SDL_mixer/PKGBUILD
+++ b/wii/SDL_mixer/PKGBUILD
@@ -14,6 +14,7 @@ depends=(
  'ppc-libvorbis'
  'ppc-libogg'
  'ppc-libmad'
+ 'ppc-libmodplug'
 )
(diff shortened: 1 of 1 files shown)


WinterMute:
Sorry, hadn't noticed this when you asked about it on Discord. The packages have been updated & necessary commits added here now. I still need to update devkitPPC package to fix an issue with building but I'll do that on the next upgrade.

PR #236, opened 2021-06-17
3ds: add wslay and nettle

Pixel-Pop:
Here is a working example program, adapted from wslay's examples, with the 3ds connecting to a websocat server: [wslay-client.zip](https://github.com/devkitPro/pacman-packages/files/6673296/wslay-client.zip)

`gcc` is in nettle's `makedepends` because it seems nettle needs to build utilities to run on the build system at build time.

Please feel free to make any changes.

### DIFF
--- a/3ds/wslay/PKGBUILD
+++ b/3ds/wslay/PKGBUILD
@@ -9,7 +9,7 @@ url='https://github.com/tatsuhiro-t/wslay'
 license=('MIT')
 options=(!strip libtool staticlibs)
 groups=('3ds-portlibs')
-makedepends=('3ds-pkg-config' 'devkitpro-pkgbuild-helpers')
+makedepends=('3ds-pkg-config' 'dkp-toolchain-vars')
 source=("https://github.com/tatsuhiro-t/wslay/releases/download/release-${pkgver}/wslay-${pkgver}.tar.gz")
 sha256sums=('90CE68C6DFD614722D44FBB14563A3F6DACC68B548B20AE382AC4F4952C55268')
 build() {
--- a/3ds/wslay/PKGBUILD
+++ b/3ds/wslay/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+
+pkgname=3ds-wslay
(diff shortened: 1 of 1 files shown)


3ds/nettle/PKGBUILD
```diff
@@ -0,0 +1,38 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+
+pkgname=3ds-nettle
```
WinterMute, line 3: Just remove the nettle package, it's not necessary for the wslay library. It's only used for examples which won't be built if nettle isn't installed.

3ds/wslay/PKGBUILD
```diff
@@ -0,0 +1,36 @@
...
+pkgname=3ds-wslay
+pkgver=1.1.1
+pkgrel=1
+pkgdesc='Wslay is a WebSocket library written in C.'
+arch=('any')
+url='https://github.com/tatsuhiro-t/wslay'
+license=('MIT')
+options=(!strip libtool staticlibs)
+groups=('3ds-portlibs')
+makedepends=('3ds-pkg-config' 'devkitpro-pkgbuild-helpers')
```
WinterMute, line 12: devkitpro-pkgbuild-helpers has been superceded by dkp-toolchain-vars and several other packages that aren't needed for this library. Just use dkp-toolchain-vars here.

WinterMute (changes requested):

WinterMute:
Don't worry about merging upstream into your fork when doing PRs, this can end up causing conflicts. If nothing you're editing has changed then you can just ignore it, otherwise it's better to rebase from upstream instead.

Pixel-Pop:
Thank you for your feedback. Should I open a separate PR for nettle?

WinterMute (approved):

WinterMute:
> Thank you for your feedback. Should I open a separate PR for nettle?

Thanks for making the changes so quickly. I've merged and packaged this now. 

I'm not sure nettle is really a good fit for 3DS unless there's something specific you're using it for that can't be done with mbedtls for instance. If nettle is added though then this package will need modding to avoid building examples.

PR #241, opened 2021-06-22
cmake+switch: create Switch specific FindOpenGL cmake module

madebr:
Using the Switch cmake toolchain, doing
```cmake
find_package(OpenGL REQUIRED)
```
fails with the following error:
```
CMake Error at /usr/share/cmake-3.13/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenGL (missing: OPENGL_opengl_LIBRARY OPENGL_glx_LIBRARY)
Call Stack (most recent call first):
  /usr/share/cmake-3.13/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.13/Modules/FindOpenGL.cmake:395 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/CMakeLists.txt:168 (find_package)
```
By re-implementing `FindOpenGL.cmake` [according to the API defined in the CMake documentation](https://cmake.org/cmake/help/latest/module/FindOpenGL.html),
`find_package(OpenGL REQUIRED)` works again.

This allows e.g. the following:
```cmake
    set(OpenGL_GL_PREFERENCE GLVND)
    find_package(OpenGL REQUIRED)

    if(TARGET OpenGL::OpenGL)
        target_link_libraries(librw
            PRIVATE
                OpenGL::OpenGL
        )
    else()
        target_link_libraries(librw
            PRIVATE
                OpenGL::GL
        )
    endif()
```


cmake/switch/Switch.cmake
```diff
@@ -14,8 +14,9 @@ set(CMAKE_ASM_FLAGS_INIT "${NX_COMMON_FLAGS}")
 
 set(CMAKE_EXE_LINKER_FLAGS_INIT "${NX_ARCH_SETTINGS} ${NX_LIB_DIRS} -fPIE -specs=${DEVKITPRO}/libnx/switch.specs")
 
-set(CMAKE_FIND_ROOT_PATH
-	${CMAKE_FIND_ROOT_PATH}
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/switch")
```
madebr, line 17: This is the only thing I'm not so happy about because [the documentation of `CMAKE_MODULE_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_MODULE_PATH.html) says that this variable is intented for projects.

WinterMute:
I'm not sure this belongs here. I think it may belong with mesa which supplies our gl libraries & it should probably be installed in `/opt/devkitpro/portlibs/switch/cmake` or maybe `/opt/devkitpro/portlibs/switch/lib/cmake`.

madebr:
I'm asking around in the #cmake channel on cpplang.slack.
They suggest using cache variables and keep using CMake's `FindOpenGL.cmake`.

Adding
```cmake
set(OPENGL_opengl_LIBRARY "${DEVKITPRO}/portlibs/switch/lib/libEGL.a" CACHE)
```
makes `find_package(OpenGL REQUIRED EGL)` succeed.

But actually using it fails because of missing symbols (the switch specific `glapi` and `drm_nouveau` dependencies are not used by CMake's find_package).

Both of the following fail also:
```cmake
# Use ';' separator: linking fails because it is interpreted as one file
set(OPENGL_opengl_LIBRARY "${DEVKITPRO}/portlibs/switch/lib/libEGL.a;${DEVKITPRO}/portlibs/switch/lib/libglapi.a;${DEVKITPRO}/portlibs/switch/lib/libdrm_nouveau.a" CACHE PATH "Switch OpenGL libraries + dependencies.")

# Use ' ' separator: the build system (gnu make) fails because it interprets the files as targets.
set(OPENGL_opengl_LIBRARY "${DEVKITPRO}/portlibs/switch/lib/libEGL.a ${DEVKITPRO}/portlibs/switch/lib/libglapi.a ${DEVKITPRO}/portlibs/switch/lib/libdrm_nouveau.a" CACHE PATH "Switch OpenGL libraries + dependencies.")
```
Using cache variables in combination with `FindOpenGL.cmake` only supports one file location.

If devkit pro would provide a library that merges `libEGL.a`, `libglapi.a` and `libdrm_nouveau.a`, then cmake could be made happy with only one library:
```
${DEVKITPRO}/devkitA64/bin/aarch64-none-elf-ar crsT libGL.a ${DEVKITPRO}/portlibs/switch/lib/libEGL.a ${DEVKITPRO}/portlibs/switch/lib/libdrm_nouveau.a ${DEVKITPRO}/portlibs/switch/lib/libglapi.a
```
But I think this is not a nice solution.

> I'm not sure this belongs here. I think it may belong with mesa which supplies our gl libraries & it should probably be installed in `/opt/devkitpro/portlibs/switch/cmake` or maybe `/opt/devkitpro/portlibs/switch/lib/cmake`. If I'm reading the docs right then that should remove the need to modify CMAKE_MODULE_PATH

Only installing it to `/opt/devkitpro/portlibs/switch/lib/cmake` does not let cmake use/find it.
I also tried using `CMAKE_SYSTEM_PREFIX_PATH`, but no success.
Only `CMAKE_MODULE_PATH` gave success.

WinterMute:
After a lot of fiddling around and headscratching we found what seems a cleaner way to do this with the caveat that it requires cmake 3.15 or later. We took your FindOpenGL.cmake and added it to the mesa package as OpenGLConfig.cmake in https://github.com/devkitPro/pacman-packages/commit/3119ed5c20b1aa897dc31d01b58bf6db8b98e567 & setting CMAKE_FIND_PACKAGE_PREFER_CONFIG in Switch.cmake at https://github.com/devkitPro/pacman-packages/commit/d9f8148c77c09023e1bffb333ef125c40307e8c8 then find_package(OpenGL REQUIRED) works as expected.

I've updated and released our packages so hopefully this should just work for you now.

PR #247, opened 2021-07-20
3ds: added gettext

tbeloqui:


mtheall:
If you are interested in gettext I wrote a lightweight implementation
https://github.com/mtheall/ftpd/blob/feature/i18n/source/gettext.cpp

PR #248, opened 2021-07-20
3ds: added libiconv

tbeloqui:


mtheall:
What's this needed for? This is a pretty heavy dependency, and it would be awkward to support the circular dependency between libiconv and gettext. If you only need utf8/utf16/utf32, libctru has lightweight conversion code for those.

tbeloqui:
@mtheall thanks for the feedback. I initially added this one as it as dependency for glib (which is a work in progress port) but if libcrtu has conversion functions for this I might then relay on those rather than pulling up the dependency.

WinterMute:
> @mtheall thanks for the feedback. I initially added this one as it as dependency for glib (which is a work in progress port) but if libcrtu has conversion functions for this I might then relay on those rather than pulling up the dependency.

Being honest glib is a pretty hefty dependency as well. What's the ultimate goal here? i.e. what is it you're intending to port that needs these things?

tbeloqui:
Yes I understand that, so I am ok to close the PR if the community wont benefit from this. The ultimate goal is to port gstreamer.

PR #252, opened 2021-08-01
Fix some SDL packages with misleading descriptions

sulix:
The Gamecube SDL and SDL_ttf packages, and the Wii SDL_ttf package have descriptions which refer to the wrong console. 

This confused me for a second, so here's a fix.


WinterMute:
I think it's probably better to keep the console name out of the descriptions going forward & rely on the package name instead. I've gone with that change for these libraries instead.

PR #254, opened 2021-08-31
Some cmake touchups

carstene1ns:


fincs:
A lot of what's in this PR has now been integrated into devkitPro CMake.

Ghabry:
@fincs yeah this looks quite complete but could you also add the change in this commit?

 https://github.com/devkitPro/pacman-packages/pull/254/commits/efd26177a09e0d8c37ab03638329d66756a441dc
 (Allow CMake to look in other directories).

By default this does not change the behaviour and is useful to optionally add additional paths were libraries are searched. For testing etc. Other toolchains use similar code.

fincs:
The behavior introduced in that commit doesn't seem to match the description, unless I'm mistaken.

> This variable controls whether the [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) and [CMAKE_SYSROOT](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSROOT.html#variable:CMAKE_SYSROOT) are used by [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package).

> 
> If set to ONLY, then only the roots in [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) will be searched. If set to NEVER, then the roots in [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) will be ignored and only the host system root will be used. If set to BOTH, then the host system paths and the paths in [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) will be searched.

`CMAKE_SYSROOT`/`--sysroot` are not used with devkitPro toolchains.

carstene1ns:
In this case is not about sysroots, just about additional library directories like mentioned in the description. Setting `CMAKE_PREFIX_PATH` did not have effect before.

However, I guess we may also use other workarounds.

fincs:
How is that supposed to work anyway? I don't see any mention of CMAKE_PREFIX_PATH anywhere except in the documentation of the find_**() family of functions.

Ghabry:
CMake does not have the best documentation.

``CMAKE_PREFIX_PATH`` is appended to all values in ``CMAKE_FIND_ROOT_PATH`` and then these directories are searched.
``CMAKE_FIND_ROOT_PATH`` is intended to be set by the toolchain file and ``CMAKE_PREFIX_PATH`` by the developer invoking CMake.

Lets say I have an additional library that is not bundled with devkitpro in ``/home/gab/switch`` then ``-DCMAKE_PREFIX_PATH=/home/gab/switch`` will not find the library but in combination with ``-DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=BOTH -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=BOTH -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=BOTH`` it will work because then it also searched from the filesystem root.

(Yeah i know that this looks terrible syntax wise but CMake is quite verbose :) )

fincs:
I think I see what's going on here.

>The CMake variable [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/v3.22/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) specifies one or more directories to be prepended to all other search directories. This effectively "re-roots" the entire search under given locations. Paths which are descendants of the [CMAKE_STAGING_PREFIX](https://cmake.org/cmake/help/v3.22/variable/CMAKE_STAGING_PREFIX.html#variable:CMAKE_STAGING_PREFIX) are excluded from this re-rooting, because that variable is always a path on the host system. By default the [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/v3.22/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) is empty.

This is what we're using right now. We set `CMAKE_FIND_ROOT_PATH` to a list of directories that have the typical bin/include/lib folder structure inside. For example, on Switch one of these dirs is `/opt/devkitpro/libnx`. All of these directories are being used as virtual filesystem roots, and all other directories that are to be searched are appended to each of these roots.

>The [CMAKE_SYSROOT](https://cmake.org/cmake/help/v3.22/variable/CMAKE_SYSROOT.html#variable:CMAKE_SYSROOT) variable can also be used to specify exactly one directory to use as a prefix. Setting [CMAKE_SYSROOT](https://cmake.org/cmake/help/v3.22/variable/CMAKE_SYSROOT.html#variable:CMAKE_SYSROOT) also has other effects. See the documentation for that variable for more.
>
> (...)
>
> Path to pass to the compiler in the --sysroot flag.
>
>The CMAKE_SYSROOT content is passed to the compiler in the --sysroot flag, if supported. The path is also stripped from the RPATH/RUNPATH if necessary on installation. The CMAKE_SYSROOT is also used to prefix paths searched by the find_* commands.
>
>This variable may only be set in a toolchain file specified by the [CMAKE_TOOLCHAIN_FILE](https://cmake.org/cmake/help/v3.22/variable/CMAKE_TOOLCHAIN_FILE.html#variable:CMAKE_TOOLCHAIN_FILE) variable.

It looks like by default sysroot is effectively `/`, i.e. the root directory. Enabling usage of this root by changing the root path mode may "fix" CMAKE_PREFIX_PATH (and friends), but it comes at the cost of CMake now searching in the host's native include/lib directories, which is extremely undesirable behavior.

One of the main approaches and goals we're attempting to follow here is avoiding monkey-see monkey-do situations that seem so incredibly common in CMake machinery. Anything that we adopt needs to be properly understood and justified, and the best possible/correct solution should be selected in order to minimize the possibility of undesired side effects.

I am going to continue digging into this in order to see if it's possible to provide a solution that does not introduce said undesired side effects. With that said, people are strongly encouraged to contribute portlibs directly to [pacman-packages](https://github.com/devkitPro/pacman-packages) in order to eliminate the "need" for third party prefixes altogether.

Ghabry:
Yeah we are aware that custom prefixes are discouraged. Thank you for looking into this.
When possible the portlib libraries are used by us.

In our project (EasyRPG Player) we cannot fully avoid custom libs:

We have the problem that we need ICU for proper Unicode support (iconv lacks some features we need) and by default ICU is 20 MB. We use a custom icudata library that is trimmed down to suite exactly our needs to make it 3 MB.
So this is unfortunately useless for general purpose and is a compiletime setting :/

Recently we also added fluidsynth for playing SF2 files. Unfortunately it depends on glib (good luck compiling this...) and pthreads (not supported on all platforms) so we removed glib and stubbed the threading.
It works for us but not sure if anybody wants this hacky thing.

fincs:
A solution for the `CMAKE_PREFIX_PATH` problem that works on CMake 3.16+ has been found, and implemented. Previous versions of CMake (from 3.13 onwards) are still supported, but they keep the old behavior.

Ghabry:
Very cool. Thanks!

This also appears to have fixed an issue I was planning to report because ``find_package(OpenGL REQUIRED)`` was not working because CMake did not correctly look for config files in portlibs.

This solves all roadblocks we had :)

PR #269, opened 2022-03-19
harfbuzz: Apply toolchain vars

rhaleblian:


WinterMute:
We don't need to do this for configuring & building with our meson wrapper, it's handled internally. Last time I checked meson doesn't pay attention to environment variables at all.

Did you have an issue that this seemed to fix?

rhaleblian:
Yep, a build a unanointed fresh shell breaks without it.
Is this some preamble or context I am ignoring?

PR #270, opened 2022-04-01
switch-libtheora: upgrade to 1.2.0alpha1

TurtleP:
This pull request upgrades switch-libtheora to 1.2.0alpha1. It puts it in line with the 3ds version. It builds and works on-console. Originally I was using switch-libtheora1.1, but this was crashing on `th_decode_packetin`. This actually fixes that. Although I cannot be sure that maybe the old 1.1 package was built incorrectly, for example.

### DIFF
--- a/switch/libtheora/PKGBUILD
+++ b/switch/libtheora/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=switch-libtheora
 pkgver=1.2.0alpha1
-pkgrel=3
+pkgrel=1
 pkgdesc='Free and open video compression codec from the Xiph.org Foundation'
 arch=('any')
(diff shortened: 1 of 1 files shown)


switch/libtheora/PKGBUILD
```diff
@@ -1,18 +1,18 @@
 # Maintainer:  Dave Murphy <davem@devkitpro.org>
 
 pkgname=switch-libtheora
-pkgver=1.1.1
-pkgrel=2
+pkgver=1.2.0alpha1
+pkgrel=3
```
oreo639, line 5: pkgrel should be set to 1 when pkgver is changed

TurtleP:
fixed

PR #271, opened 2022-04-03
add libfaad2

badda71:
Freeware Advanced Audio (AAC) Decoder including SBR decoding


badda71:
library compiles fine but segfaults :-(

PR #275, opened 2022-07-22
Add devkitarm-rules to makedepends of devkitarm-crtls

Rafflesiaceae:
During the build phase of devkitarm-crtls the 'base_rules' of the
toolchain, provided by devkitarm-rules, are required.

Therefore devkitarm-rules must've been installed prior.


WinterMute:
devkitarm-rules is installed by devkitARM as a dependency

PR #285, opened 2022-10-01
cmake: Make platform ID flags cached

Nosh-Ware:
I found that there wasn't any easy type of define given for platform flags. When coding a multi platform header, it would make things a lot simpler if a define like NINTENDO_WII was made so somebody could use macros to determine dependencies or platform specific functionality. Ex.

```
#ifdef NINTENDO_WII
   #include <gccore.h>
#else
   #include <cstdint> 
   typedef int8_t s8;
#endif
```

### DIFF
--- a/cmake/3ds/Nintendo3DS.cmake
+++ b/cmake/3ds/Nintendo3DS.cmake
@@ -8,7 +8,7 @@ include_guard(GLOBAL)
 include(Platform/Generic-dkP)
 
 # Platform identification flags
-set(NINTENDO_3DS TRUE)
+set(NINTENDO_3DS CACHE BOOL TRUE)
 
 # Platform settings
 set(CTR_ARCH_SETTINGS "-march=armv6k -mtune=mpcore -mfloat-abi=hard -mtp=soft")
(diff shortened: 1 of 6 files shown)


fincs:
What is this PR intended to solve? These variables are supposed to be used like standard CMake variables `if(WIN32)`, `if(UNIX)`, `if(APPLE)`, etc; that is, `if(NINTENDO_3DS)`, `if(NINTENDO_SWITCH)`, etc.

Nosh-Ware:
It can be useful to have these available to the user as they my need platform specific definitions or includes etc. 
Just added an example to the main message above.

fincs:
Those are completely unrelated to CMake variables. You want C/C++ preprocessor defines. Each platform already has such defines, e.g. `__3DS__`, `__SWITCH__`. However you brought into my attention the fact that no preprocessor defines are currently added to distinguish GameCube from Wii. I'll see what I can do about that.

Nosh-Ware:
Alright, well, happy to help, even tangentially.

fincs:
Update: Turns out we do in fact provide preprocessor defines for Wii, namely `__wii__` (lowercase) and `HW_RVL`. GameCube on the other hand is `__gamecube__` and `HW_DOL`. These are automatically defined by GCC when selecting GC/Wii compilation mode, so it's not part of CMake logic.

PR #305, opened 2023-07-06
Updated SDL2, SDL2_image, SDL2_mixer and SDL2_ttf

giroletm:
# Updated SDL2 and three of its variants
- SDL2: 2.0.14 -> 2.28.1
- SDL2_image: 2.0.4 -> 2.6.3
- SDL2_mixer: 2.0.4 -> 2.6.3
- SDL2_ttf: 2.0.15 -> 2.20.2

# Publishing
I originally wanted to make a post on the dkp forums to ask this, but there's apparently an internal server error.

I'd like to make a pull request to send my contributions to https://github.com/devkitPro/SDL, but due to the way the repo is organized, I don't really see how to.

I could sure send a pull request to the ```switch-sdl2``` branch, but it'd result in an enormous commit that wouldn't mean much when looking at the diff.

The best would be to make a ``switch-sdl2-2.28.1`` branch (which would make sense since there is a ``switch-sdl2-2.0.14`` branch), which would be based on SDL's ``release-2.28.1`` branch and have my changes commited on top of it. The ```switch-sdl2``` would then base itself on the new ``switch-sdl2-2.28.1``. This would keep the previous history in the old 2.0.14 branch and the new one in the new branches.

I did exactly this here: https://github.com/giroletm/dkp-SDL/tree/switch-sdl2-2.28.1

The thing is, you can't send "branch creation requests" so I don't know how am I supposed to send these. I don't want someone else to commit my changes for me, since I'd like to have my git identity as the commit's author for crediting reasons that seem reasonable to me.

(this would also require the creation of a fork of the original ``SDL_ttf`` repo, because I have added edits to it to make it switch-compatible, while the original switch portlib has no edits for it whatsoever)

So please, tell me what to do on that side.

# Building
All three build successfully, although you need to use some tricks (that aren't the fault of my update but ``makepkg``).

If needed, I can provide builds.

## Allowing ``curl`` to request files from ``libsdl.org``
Open the ``devkitPro/msys2/etc/makepkg.conf`` file (with ``devkitPro being your dkp installation folder) in a text editor

On line 14, change ``'https::/usr/bin/curl -gqb "" -fLC - --retry 3 --retry-delay 3 -o %o %u'`` to ``'https::/usr/bin/curl -gqb "" -fLC - --retry 3 --retry-delay 3 -o %o %u --insecure'``.

This will make curl skip ssl verification for https requests, allowing it to download from ``libsdl.org``, since its certificate isn't part of dkp's curl installation.

## Symbolic links issue

The downloaded ``tar.gz`` file contains symbolic links, which cause a ``bsdtar`` error when running ``makepkg``.
This happens with SDL2_image, SDL2_mixer and SDL2_ttf:
```
==> Extracting sources...
  -> Extracting SDL2_ttf-2.20.2.tar.gz with bsdtar
SDL2_ttf-2.20.2/Xcode/macOS/SDL2.framework/Resources: Can't replace existing directory with non-directory: Directory not empty
SDL2_ttf-2.20.2/Xcode/macOS/SDL2.framework/Headers: Can't replace existing directory with non-directory: Directory not empty
SDL2_ttf-2.20.2/Xcode/macOS/SDL2.framework/Versions/Current: Can't replace existing directory with non-directory: Directory not empty
bsdtar: Error exit delayed from previous errors.
==> ERROR: Failed to extract SDL2_ttf-2.20.2.tar.gz
    Aborting...
```
*Example with SDL2_ttf*

This is due to ``bsdtar`` not being able to extract Linux-like symbolic links on windows.

Here is the way I worked around it, but if there is a ``PKGBuild`` setting to just ignore ``bsdtar`` errors, please tell me.
- Run ``makepkg`` and see the error (this is required)
- Open WinRAR (or another tool capable of unpacking ``tar.gz`` archives) in **Administrator mode**
- In WinRAR, navigate to ``pacman-packages\switch\SDL2_THEMODULEYOUWANT\src`` and delete the ``SDL2_THEMODULEYOUWANT-VERSIONNUMBER`` **directory** (replace SDL2_THEMODULEYOUWANT with the SDL2 module of your choice, and VERSIONNUMBER with its version number)
- Right click on the ``SDL2_THEMODULEYOUWANT-VERSIONNUMBER.tar.gz`` **file** and select "Extract without confirmation"
- Run ``makepkg`` again

This way, WinRAR handles the symbolic links and creates them in a way that seem to satisfy ``makepkg``.

### DIFF
--- a/switch/SDL2/PKGBUILD
+++ b/switch/SDL2/PKGBUILD
@@ -1,8 +1,9 @@
 # Maintainer:  Dave Murphy <davem@devkitpro.org>
+# Contributor: giroletm <rsmromaji@gmail.com>
 
 pkgname=switch-sdl2
-pkgver=2.0.14
-pkgrel=2
+pkgver=2.28.1
+pkgrel=1
 pkgdesc="A library for portable low-level access to a video framebuffer, audio output, mouse, and keyboard (Nintendo Switch port)"
 arch=('any')
 url="https://libsdl.org"
@@ -44,5 +45,5 @@ package() {
   # remove useless stuff
   rm -r "$pkgdir"${PORTLIBS_PREFIX}/share
 }
-sha256sums=('d8215b571a581be1332d2106f8036fcb03d12a70bae01e20f424976d275432bc'
-            'fc2cf84907394b7d988697483b7b726c90eed3c28f982799555daa44490038d2')
+sha256sums=('4977ceba5c0054dbe6c2f114641aced43ce3bf2b41ea64b6a372d6ba129cb15d'
+            'f3cd18a86c50497d5539e372486d87907e51ad57ba68e1c89deebfcb11873bf9')
--- a/switch/SDL2/SDL2-2.28.1.patch
+++ b/switch/SDL2/SDL2-2.28.1.patch
@@ -1,194 +1,20 @@
-diff --git a/.gitignore b/.gitignore
-new file mode 100644
-index 000000000..7f5da9cd6
---- /dev/null
-+++ b/.gitignore
-@@ -0,0 +1,168 @@
-+aclocal.m4
-+autom4te*
-+config.cache
-+config.log
-+config.status
-+libtool
-+Makefile
-+Makefile.rules
-+sdl2-config
-+sdl2-config.cmake
-+sdl2-config-version.cmake
-+sdl2.pc
-+SDL2.spec
-+build
-+gen
-+Build
-+buildbot
-+
-+*.so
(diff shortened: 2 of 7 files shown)


WinterMute:
PRs to this repository should generally be done as a single PR per package. For packages that use patches generated from our own forks we prefer PRs to those forks. Building packages on msys is not supported, there are issues with permissions on the generated packages which cause problems on macOS & linux. SDL2_ttf can be built fine without patching.

PR #307, opened 2023-08-20
switch: add libdav1d, update FFmpeg with hwaccel

averne:
As discussed in https://github.com/switchbrew/libnx/issues/580. 
Since FFmpeg does not implement its own AV1 decoder, I also added dav1d, which I think is the fastest software implementation as of now.


averne:
Tested working on my (unreleased) player:

https://github.com/devkitPro/pacman-packages/assets/45773016/d6c02472-e0af-4413-b6b0-de7864055992

WinterMute:
Good job. Thanks for doing this.

PR #312, opened 2023-09-12
switch: add libavif and switch: sdl2_image 2.0.4 -> 2.6.3

Otorhin:


WinterMute:
One package per PR please.

We also want to look at updating the SDL2 package before updating packages that depend on it.

PR #314, opened 2023-09-30
Fix Generic-dkP path issue

wally4000:
cmake is not able to find the Generic-dkP file the way it is currently :)


fincs:
This is a package made out of loose files. The PKGBUILD [installs them to their correct locations](https://github.com/devkitPro/pacman-packages/blob/master/cmake/3ds/PKGBUILD#L27-L30).

If you want to use devkitPro CMake, you only need to install the appropriate packages using pacman: `pacman -S 3ds-cmake`

PR #318, opened 2023-10-11
switch: Add objfw

Midar:
This adds objfw for switch to allow using Objective-C for Switch development.


WinterMute:
While I commend your enthusiasm and persistence with Objective C I don't really understand how this is actually useful for the users of our toolchains on our supported platforms. Can you write a game with this? How do you go about using the platform libraries we supply and the other portlibs?

Midar:
Tes, it absolutely can be used to write a game! That's the entire point of proposing to import it. Using platform libraries and portlibs is no problem either, it's just the same as with C++ - in fact, it's even more compatible as with C++ (unlike C++, it's a *strict* superset of C and not just mostly a superset of C). Basically, consider this to be "Add libsupc++, libstdc++ and boost", except for ObjC instead of C++.

Midar:
@WinterMute So, should I update this to 1.0.5 or will it be rejected?

WinterMute:
Some example code and a few more people interested in playing with Objective C would help. It's hard to know what to do with something I've never used and few people are asking for.

PR #323, opened 2023-10-23
switch-libmpv: fix vic alignement for hwaccel

dragonflylee:
support ffmpeg tx1 hwaccel


WinterMute:
Why did you close this?

PR #329, opened 2023-12-06
switch-libmpv: update to 0.37.0

dragonflylee:


WinterMute:
Could you move libplacebo to a separate PR & remove the  mingw-w64 tagging there please?

averne:
I don't mind seeing my audio backend upstreamed to the dkP package, but I would appreciate not finding out by chance... especially when this version of my patch has never been shared publicly and originates from a private repo. 
So at least ping me about it on discord or in the PR

PR #345, opened 2024-02-23
switch: add ntfs-3g

DarkMatterCore:
Based on the original PKGBUILD script + patch from the libusbhsfs repository, plus some extra modifications based on the PPC portlib.

References:

* https://github.com/devkitPro/docker/pull/28
* https://devkitpro.org/viewtopic.php?p=17868

### DIFF
--- a/switch/ntfs-3g/PKGBUILD
+++ b/switch/ntfs-3g/PKGBUILD
@@ -1,0 +1,58 @@
+# Maintainer: Rhys Koedijk <rhys@koedijk.co.nz>
+# Maintainer: DarkMatterCore <pabloacurielz@gmail.com>
+
+pkgname=switch-ntfs-3g
+pkgver=2022.10.3
+pkgrel=1
+pkgdesc='An open source, freely available read/write NTFS driver'
+arch=('any')
+url='https://www.tuxera.com/community/open-source-ntfs-3g/'
+license=('GPL')
+options=(!strip libtool staticlibs)
+groups=('switch-portlibs')
+makedepends=('devkitA64' 'dkp-toolchain-vars')
+
+source=(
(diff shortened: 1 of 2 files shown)


DarkMatterCore:
It seems like the package generated by this build script isn't getting fully populated. After running `pacman -S switch-ntfs-3g` and checking `/opt/devkitpro/portlibs/switch`, only the NTFS-3G license is getting installed.

I'll look into the issue.

WinterMute:
Heh, I didn't check the package contents, sorry. The PKGBUILD was missing the install step & it looks like that bit of the patch I thought wasn't necessary probably was. I've rebuilt the package, should be fine now.

DarkMatterCore:
> Heh, I didn't check the package contents, sorry. The PKGBUILD was missing the install step & it looks like that bit of the patch I thought wasn't necessary probably was. I've rebuilt the package, should be fine now.

Thanks! I appreciate it.

Btw, I also noticed we're not copying the `config.h` file that gets generated during the NTFS-3G build steps. I created an additional PR to address that: https://github.com/devkitPro/pacman-packages/pull/347

Hope that helps.

averne:
@WinterMute Would you be accepting PKGBUILDs for libnfs and libsmb2 as well? I have those on my tree but never tried to upstream them, since the team's standing seemed to be that filesystem implementations should live in sysmodules.

In the case of libsmb2, there is also an [old PR](https://github.com/devkitPro/pacman-packages/pull/116) that went inactive since `readv` and `writev` are not provided by newlib. Atomicity can be achieved by just copying the data to a single buffer and sending that over. Not pretty of course but this is what [glibc does](https://github.com/bminor/glibc/blob/master/sysdeps/posix/readv.c).

PR #347, opened 2024-02-23
switch-ntfs-3g: install config.h

DarkMatterCore:


WinterMute:
That's really odd. You'd think make install would take care of this.

While digging around upstream issues I found this which amused me https://github.com/tuxera/ntfs-3g/issues/80#issuecomment-1860861654

DarkMatterCore:
It's strange, yeah. It's been a good three years or so since I last messed the NTFS-3G portlib (besides keeping it up to date), so I didn't really remember why we implemented those seemingly weird changes to both the PKGBUILD script and the diff patch. I guess we've got our answer.

In any case, thanks again!

WinterMute:
> In any case, thanks again!

No worries.

I made another couple of tweaks to use the official release tarball which avoids needing the autogen & used sed to get rid of install-exec-hook instead of patching Makefile.am. PR'd the fixes upstream @ https://github.com/tuxera/ntfs-3g/pull/105 so maybe that'll get merged for next release.

PR #366, opened 2024-10-26
switch-ffmpeg: update to 7.1

averne:
This moves ffmpeg to its current latest version, with an updated version of the hwaccel patch.
The hwaccel backend has been renamed from tx1 to nvtegra, with some additions (notably vp9 backward updates) and numerous bugfixes.
Tested working on my media player.


WinterMute:
Thanks for the PR, much appreciated.

I neglected to check pkgrel before I merged this & rolled a package so it's too late to fix it for this one but pkgrel should be reset to 1 when the version changes. It's not a big deal.

averne:
> I neglected to check pkgrel before I merged this & rolled a package so it's too late to fix it for this one but pkgrel should be reset to 1 when the version changes. It's not a big deal.

Ah sorry, missed this.

PR #370, opened 2024-12-04
Basic port of curl and mbedtls to NDS

ds-sloth:
This is a bit of a "feature request" PR. I've made a very basic port of mbedtls and curl to dswifi / calico, and I'd love for this port to evolve into something well-supported and widely used, but I don't think that I personally have all of the skills needed to do so.

There may be some issues with my current implementation and needs more testing to be called "stable", but it may be of use to the community.

Here are the "interesting" features of my ports:

**mbedtls**:
* Disables the SSL server protocol
* Disables IPv6 and `getaddrinfo` (both of which are not included in functions called by libcurl
* Uses `recv` / `send` instead of `read` / `write` (I believe that the sgIP stack does not use a unified namespace for sockets and file descriptors?)

**curl**:
* Disables all protocols other than HTTP/HTTPS, FTP/FTPS, and FILE
* Uses `ioctl` instead of `sfcntl` (for same reason as above)
* Uses `closesocket` instead of `close` (by putting `${DEVKITPRO}/libnds/include/sys/` into the include path so that `socket.h` is detected)

Note that I listed Wintermute as a "maintainer" for the purposes of this PR but I am happy to change that if you like.

Finally note that these versions are both out of date. Unfortunately, current versions don't seem to work properly and I am not certain why. (Newer versions also increase the codesize significantly.)

### DIFF
--- a/nds/curl/PKGBUILD
+++ b/nds/curl/PKGBUILD
@@ -1,0 +1,85 @@
+# Maintainer: WinterMute <davem@devkitpro.org>
+# Contributor: ds-sloth
+
+pkgname=nds-curl
+pkgver=7.83.0
+pkgrel=1
+pkgdesc='This build includes only the HTTP/HTTPS, FTP/FTPS, and FILE protocols for codesize.'
+arch=('any')
+url='https://curl.haxx.se/'
+license=('MIT')
+options=(!strip libtool staticlibs)
+depends=(
+         "nds-mbedtls"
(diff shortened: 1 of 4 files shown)


WinterMute:
We have some of this in progress already but we intend to further update the sgIP layer to hook into newlib rather than patching curl and mbedtls for shortcomings there. I'd prefer not to have workarounds in these packages, sorry.

PR #374, opened 2025-02-09
wii-sdl2_net: new package

mardy:
SDL2_net has had tons of commits since the last tagged release, so I'm picking the commit hash here. Please let me know if I should do otherwise.

If you create a repository for SDL_net, I will push my two patches there. But it's not essential, as I don't think we'll be making more changes to this package.

I will submit a version for the GameCube after this package gets reviewed (unfortunately we need a different package for the GameCube, since we use net_select for the GameCube and net_poll for the Wii.

### DIFF
--- a/wii/SDL2_net/PKGBUILD
+++ b/wii/SDL2_net/PKGBUILD
@@ -1,0 +1,39 @@
+# Maintainer:  Dave Murphy <davem@devkitpro.org>
+# Contributor: Alberto Mardegan <mardy@users.sourceforge.net>
+
+_realname=sdl2_net
+_commit="da64344c8a48d49a764833fc46f691a21115887e"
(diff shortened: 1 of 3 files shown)


Fancy2209:
Now that net_select has been added to Wii, shouldn't this be changed so it's one package for both?

mardy:
> Now that net_select has been added to Wii, shouldn't this be changed so it's one package for both?

If it was possible, yes, but I think that since there are two versions of libogc, it's impossible (or just not natural) for the pacman repository to provide only a single version of sdl2-net. But if @WinterMute contradicts me, then I'll happily psuh a PR for the ppc repo.

WinterMute:
Would prefer this not be done without completing porting work on standard berkeley sockets will email you about what we have already Alberto.

PR #376, opened 2025-03-25
Use `CMAKE_TOOLCHAIN_FILE` env var instead of passing as `-D` parameter

capitalistspz:
The `-D` option can't be used with some CMake subcommands, including `--build` and `--install`. 

Solves CLion not being able to build projects when the toolchain CMake path is set to one of the `<platform>-cmake` scripts.


WinterMute:
Thanks for that, much appreciated. 

I'd been considering doing something with the wrappers to detect --build & --install but the necessary circular tuits have been in short supply.

PR #379, opened 2025-05-14
switch-mbedtls: Workaround to fix SSL error because of gcc15

dragonflylee:
refer https://gitlab.linphone.org/BC/public/linphone-sdk/-/merge_requests/5026


dragonflylee:
2.28.10 was the last release of 2.28.x of mbedtls

PR #387, opened 2025-07-27
feat(3ds-curl): enable websockets

gradylink:
This only adds 7 kilobytes to the 3ds-curl package which is very little compared to the original 422 kilobytes before hand. Adding 7 kilobytes should not have very much effect on many applications. If you want I can turn this into it's own package.


gradylink:
Is there a reason this hasn't been merged? It's pretty annoying for our project to rely on installing a seperate package.

gradylink:
Replaced by #423.

PR #399, opened 2025-12-18
Add Libsmb2

Wolf3s:


Wolf3s:
> * We build individual platform packages from their own separate PKGBUILDs
> * readv and writev don't belong in this package
> * we only build from stable release tags, not from git head

Ok til next time then.

PR #409, opened 2026-02-04
PPC: add ppc-libzstd package.

dkosmari:
This adds the ppc-libzstd package.

I'm not sure if the package should be called `ppc-zstd` instead.


dkosmari:
On the other hand we're not providing the full zstd (or brotli) package, only the library. It's common for distros to split the tools from the libraries. That's why I went with the `lib` prefix in the package name. But I have no preference either way.

PR #418, opened 2026-02-12
PPC: update ppc-harfbuzz to 12.3.2

dkosmari:
This updates ppc-harfbuzz to harfbuzz 12.3.2.

~~Since they changed the build system to meson, it needs a bit of patching to disable `pic`.~~


dkosmari:
Dropped the patch since I Extrems told me about the `-Db_staticpic=false` option.
Could that be set by default in `meson-toolchain.sh`?

PR #423, opened 2026-03-17
feat(3ds-curl): update to 8.15.0

gradylink:
This also updates the URL the `3ds-curl` package pulls from, switches the build to use CMake instead of autotools, and removes the duplicate `sha256sums`.


matu6968:
With a little bit of figuring out code path changes, i have now ported version 8.20.0...but it needs a newer MbedTLS (3.2.0+) which i have yet to upstream the update:

[libcurl.patch](https://github.com/user-attachments/files/28477999/libcurl.patch)

